PR #45, opened 2015-10-09
New methods for excluding elements with specific missing or empty attributes

forum-is:
We recently came across a problem where an **img**-element with missing **src**-attribute caused our PDF engine to break. As the current version of the html sanitizer only contains a method for disallowing elements without **any** attributes, we suggest adding methods to target a specific missing/empty attribute.
As a convenience we also included a method to check for a non-matching regex pattern.


jmanico:
Thank you very, very much for submitting a patch with your report. More from Mike soon!

Aloha,
--
Jim Manico
@Manicode

src/main/java/org/owasp/html/HtmlPolicyBuilder.java
```diff
@@ -288,7 +288,28 @@ public HtmlPolicyBuilder allowWithoutAttributes(String... elementNames) {
     }
     return this;
   }
-
+  	
+  /**
+   * Disallows the given element from appearing without the given attribute.
+   */
+  public HtmlPolicyBuilder disallowWithoutAttribute(String elementName, final String attributeName) {
```
mikesamuel: Right now, we can say

```java
    myPolicyBuilder.allowAttributes("src", ...).onElements("img")
```

and if I understand your goal, that is problematic because it allows but does not require `src="..."` on ``.

I'd prefer

```java
    myPolicyBuilder.withAttributes("src", ...).required().onElements("img")
```

which allows mixing `required()` into the existing flow by which elements are associated with attributes instead of creating a new API.

src/main/java/org/owasp/html/HtmlPolicyBuilder.java
```diff
@@ -714,6 +735,32 @@ public AttributeBuilder matching(
...
+
+    /**
+     * Restrict the values allowed by later {@code allow*} calls to those
+     * that are not only whitespace or an empty string. Can be combined 
+	 * with calls to {@code matching} or {@code notMatching}.
+     */
+    public AttributeBuilder notEmptyOrWhitespace() {
+      return matching(new AttributePolicy() {
+        public @Nullable String apply(String elementName, String attributeName, String value) {
+          return value.trim().length() > 0 ? value : null;
```
mikesamuel: I believe that the definition of whitespace in HTML is not the same as the whitespace matched by `String.trim()`.

http://www.w3.org/html/wg/drafts/html/master/single-page.html#space-character says

> The space characters, for the purposes of this specification, are U+0020 SPACE, U+0009 CHARACTER TABULATION (tab), U+000A LINE FEED (LF), U+000C FORM FEED (FF), and U+000D CARRIAGE RETURN (CR).

while [javadoc](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#trim%28%29) says

> character in the string whose code is greater than '\u0020'

src/main/java/org/owasp/html/HtmlPolicyBuilder.java
```diff
@@ -714,6 +735,32 @@ public AttributeBuilder matching(
...
+        public @Nullable String apply(String elementName, String attributeName, String value) {
+          return !pattern.matcher(value).matches() ? value : null;
+        }
+      });
+    }
+
+    /**
+     * Restrict the values allowed by later {@code allow*} calls to those
+     * that are not only whitespace or an empty string. Can be combined 
+	 * with calls to {@code matching} or {@code notMatching}.
```
mikesamuel: Please use `{@link #matching}`  instead of `{@code matching}`.  The latter allows the reader to click through and generates doc warnings should APIs change.

Please fix indentation.

Maybe add a caveat along the lines of

> Sometimes empty attributes are different from absent attributes.  For example `<input checked="">` is not equivalent to `<input>` and `<a href="">` is a link to the current document without a fragment while `<a>` is not.

mikesamuel:
This change seems to be missing any tests.  Could you at least check that it works with the kind of policy you want to work.

mikesamuel:
> We recently came across a problem where an img-element with missing src-attribute caused our PDF engine to break. As the current version of the html sanitizer only contains a method for disallowing elements without any attributes, we suggest adding methods to target a specific missing/empty attribute.
As a convenience we also included a method to check for a non-matching regex pattern.

What PDF engine is this?  If you don't maintain it then it might be worth filing a bug with them as well.

Should this change also require `src="..."` on `` in [`Sanitizers.IMAGES`](https://github.com/OWASP/java-html-sanitizer/blob/master/src/main/java/org/owasp/html/Sanitizers.java#L120)?

forum-is:
Hello Mike,

sorry for the wait.

Our goal was indeed to be able to require certain attributes when their element would make no sense or as in this case even cause harm without them, which was not possible with the current allowAttributes()-implementation.

Your proposed mimic of a chainable required() seems indeed a cleaner approach. Yet my understanding of the project code is currently not sufficient to change the implementation to that within reasonable effort. I will however try and provide a test case to show the validity of the current implementation shortly. Maybe then you could refactor it to the required()-mimic as you find the time.

We use [Apache FOP](https://xmlgraphics.apache.org/fop/) to export PDFs from our web application. 

As you mention, it might actually be useful to require the attribute in the standard sanitizer so others do not run into that kind of issue unsuspectingly. At least no valid use case comes to mind for an img without a src that would forbid it from being required.

Best regards,
Sebastian

forum-is:
Hello Mike,

I added a TestCase for the disallowWithoutAttribute() functionality that represents the constellation that caused us problems (arbitrary alt, missing src).

Best regards,
Sebastian

jmanico:
It looks like this is mostly resolved. @mikesamuel - can we merge in the test case or resolve this pull request somehow? It does not look like major work is needed to resolve this..

jmanico:
Bump @mikesamuel ?

mikesamuel:
Sorry for ghosting.  I'll take a look shortly.

jmanico:
Closing this after a long silence, with thanks. The idea, dropping an element when a specific attribute is missing or empty, is still reasonable, but this branch predates the module restructure and the rework of `allowWithoutAttributes` / `disallowWithoutAttributes` in release 20200615.1 that @mikesamuel mentioned above, so it would need to be rebuilt rather than rebased. If anyone still needs this, please open an issue describing the use case and we can discuss the right API shape against the current builder. Sorry this sat for so long, @forum-is.

Issue #52, opened 2015-12-09
<a> elements containing <div> elements are not working


bedhub:
When i sanitize links containing sub elements these elements will be moved outside the a element.
```
<a href=\"https://www.xyz.com\"><div>Button text</div></a> 
will result in:
<a href="https://www.xyz.com" rel="nofollow"></a><div>Button text</div>
```

Can you please help me with this issue? 

Example code:
```
		String html = "<a href=\"https://www.xyz.com\"><div>Button text</div></a>";
		PolicyFactory policy = new HtmlPolicyBuilder()
				.allowElements("a", "div")
				.allowUrlProtocols("https")
				.allowAttributes("href").onElements("a")
				.requireRelNofollowOnLinks()
				.toFactory();
		String safeHTML = policy.sanitize(html);
```

tyagiakhilesh:
Hello Bedhub,
Were you able to resolve this issue ? I am facing SIMILAR ( not exact ) kind of problem with table attribute. Any ideas ?

bedhub:
No, i have no solution for this. I would really appreciate if someone can look into this.

mikesamuel:
It looks like [HTML5:"The `<a>` element"](https://www.w3.org/TR/html51/semantics.html#the-a-element) has reclassified changed to use [content model *transparent*](https://www.w3.org/TR/html51/dom.html#transparent) but [`TagBalancingHtmlStreamEventReceiver`](https://github.com/OWASP/java-html-sanitizer/blob/master/src/main/java/org/owasp/html/TagBalancingHtmlStreamEventReceiver.java#L428) still classifies it as *inline*.

One way to solve this and other problems would be to implement handling for content model *transparent* and update the element tables.

nosneros:
Is there a priority list of issue that this could be added to? Considering the change reference from mikesamuel, we will be seeing this more frequently and will have unusable links.

jmanico:
Andy,
Thank you for this patch. We understand how important this is to you. Mike will take a close look as soon as he can. We'll keep this issue alive until it's fully deployed to Maven. 
Aloha + Thank you Andy!,
Jim

jmanico:
Note, this is being worked on here with further discussion on this issue. https://github.com/OWASP/java-html-sanitizer/pull/62

mikesamuel:
https://github.com/OWASP/java-html-sanitizer/tree/issue52 should fix the immediate issue.

mikesamuel:
[...]

PR #62, opened 2016-04-04
Add 'BLOCK' to <a>'s content types, so that <div> could be part of an…

andy-h-chen:
… anchor node.

This is for https://github.com/OWASP/java-html-sanitizer/issues/52

> Changes in HTML5
>Although previous versions of HTML restricted the a element to only containing phrasing content (essentially, what was in previous versions referred to as “inline” content), the a element is now transparent; that is, an instance of the a element is now allowed to also contain flow content (essentially, what was in previous versions referred to as “block” content)—if the parent element of that instance of the a element is an element that is allowed to contain flow content.

Copied https://www.w3.org/TR/html-markup/a.html

### DIFF
--- a/src/main/java/org/owasp/html/TagBalancingHtmlStreamEventReceiver.java
+++ b/src/main/java/org/owasp/html/TagBalancingHtmlStreamEventReceiver.java
@@ -429,7 +429,7 @@ public class TagBalancingHtmlStreamEventReceiver
           "a", false, elementGroupBits(
               ElementGroup.INLINE
           ), elementGroupBits(
-              ElementGroup.INLINE_MINUS_A
+              ElementGroup.BLOCK, ElementGroup.INLINE_MINUS_A
           ));
       defineElement(
           "abbr", true, elementGroupBits(


mikesamuel:
Doesn't this fix

```html
<a href=""><div>foo</div></a>
```

at the cost of breaking

```html
<span><a href=""><div>foo</div></a>
```

andy-h-chen:
@mikesamuel 
I'm testing it using SlashdotPolicyExample added 'span' to allowElements. Can I have your test app?
With my patch
```
 [Reading from STDIN]
<span><a href=""><div>foo</div></a>
<a href="" rel="nofollow"><div>foo</div></a>

[Reading from STDIN]
<a href=""><div>foo</div></a>
<a href="" rel="nofollow"><div>foo</div></a>
```
Without my patch
```
[Reading from STDIN]
<a href=""><div>foo</div></a>
<a href="" rel="nofollow"></a><div>foo</div>

[Reading from STDIN]
<span><a href=""><div>foo</div></a>
<a href="" rel="nofollow"></a><div>foo</div>
```

jmanico:
Thanks for the analysis Andy. We know this is important to you. Mike will respond as soon as he can. :) Aloha!

mikesamuel:
I think the *transparent* language in the spec affects several things.

1. `<div><a href=""><div>...</div></a></div>` does not contain any implicit end tags.
1. `<span><a href=""><span>...</span></a></span>` does not contain any implicit end tags.
3. `<span><a href=""><div>...</div></a></span>` is equivalent to `<span><a href=""></a></span><div>...</div>` unless there's some orphaning going on.

The first two validate via https://validator.w3.org/nu/#textarea when wrapped in an envelope like

```html
<!doctype html><html><head><title>.</title></head><body>
<div><a href=""><div>...</div></a></div>
</body></html>
```

and the third gives me

> **Error: Element div not allowed as child of element a in this context. (Suppressing further errors from this subtree.)**

> From line 2, column 18; to line 2, column 22

> `a href=""><div>...</d`

> Contexts in which element div may be used:

> Where flow content is expected.

> Content model for element a:

> Transparent, but there must be no interactive content descendant.

andy-h-chen:
@mikesamuel 
Thanks for the explanation. Seems the best way would be implementing transparent handling.

mikesamuel:
Yeah.  I'm surprised that this div in link thing is causing problems but since link boundary handling seems important to get right in one way I'd rather not break any existing uses of it.  Handling transparency seems the best way to balance both concerns.

mikesamuel:
Does https://github.com/OWASP/java-html-sanitizer/tree/issue52 address your problems?

andy-h-chen:
Yes, that's exactly what I'm trying to solve.

andy-h-chen:
I'll test it later. Thanks.

PR #63, opened 2016-04-09
Fixes issue 52 - block elements inside `<a>` elements

mikesamuel:
reworked TagBalancingHtmlStreamEventReceiver to handle transparency in `<a>` tags in a limited way to address issue 52.  This changes the element containment info tables to use a builder mode since the constructor calls were getting unwieldy.  It adds 4 tests, 2 of which fail currently since they are orthogonal to issue 52.  Fixing those last 2 will benefit from a larger structural change that also has the side-effect of making containment checking O(1).

### DIFF
--- a/src/main/java/org/owasp/html/TagBalancingHtmlStreamEventReceiver.java
+++ b/src/main/java/org/owasp/html/TagBalancingHtmlStreamEventReceiver.java
@@ -28,6 +28,8 @@
 
 package org.owasp.html;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.List;
 
 import javax.annotation.Nullable;
@@ -50,6 +52,9 @@ public class TagBalancingHtmlStreamEventReceiver
   private int nestingLimit = Integer.MAX_VALUE;
   private final List<ElementContainmentInfo> openElements
       = Lists.newArrayList();
+  private final Deque<ElementContainmentInfo> toResumeInReverse
+      = new ArrayDeque<ElementContainmentInfo>();
+
 
   /**
    * @param underlying An event receiver that should receive a stream of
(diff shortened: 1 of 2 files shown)


mikesamuel:
Err `s/alex/andy/`.  Sorry about that.

mikesamuel:
@andy-h-chen, Does the latest incarnation fix your problem?

andy-h-chen:
@mikesamuel , it works now. Thanks!

andy-h-chen:
@mikesamuel , just want to let you know that `<b>` needs to be updated too. Link is lost and div inside b is also invalid. See the output of SlashdotPolicyExample:

```
[Reading from STDIN]
<b><a href=''><div>def</div></a></b>
<b><a href="" rel="nofollow"></a></b><b><div>def</div></b>
```

mikesamuel:
@andy-h-chen Why do you think this is valid?

The `<div>` there is not allowed if my reading of the spec is correct.

https://validator.w3.org/nu/#textarea also reports 

> Error: Element div not allowed as child of element a in this context. (Suppressing further errors from this subtree.)

andy-h-chen:
I'm not saying it is valid but the sanitized output should be valid, right?

mikesamuel:
@andy-h-chen.  Ah I see.  The code that resumes the `<b>` that was split by the end of the `</a>` should cannot contain a `<div>`.

mikesamuel:
@andy-h-chen I think I fixed the resumption problem.  Thanks for pointing that out.

mikesamuel:
@andy-h-chen If this looks good to you, I'll prep a release.

andy-h-chen:
Looks good. Thanks!

Issue #77, opened 2016-07-07
<table> tag closed unexpectedly


andy-h-chen:
`<table style='border: solid 1px; width: 600px;'>
  <tbody>
    <tr>
      <td>
        <table style='border: solid 1px; width: 600px;'>
          <tbody>
            <tr>
              <td>1st td</td>
            </tr>
            <td>2nd td</td>
            </tr>
          </tbody>
        </table>
      </td>
    </tr>
  </tbody>
</table>`
A tr is missing before '2nd td'. Chrome browser/webview automatically adds tr before it, but html sanitizer closes table before it, and '2nd td' becomes another td of outer table.
Html sanitizer output:
`<table style="border:solid 1px;width:600px"><tbody></tbody><tbody><tr><td>
        <table style="border:solid 1px;width:600px"><tbody><tr><td>1st td</td></tr></tbody></table></td><td>2nd td</td></tr></tbody></table>`

andy-h-chen:
https://html.spec.whatwg.org/multipage/syntax.html#parsing-main-intbody

> 

    When the user agent is to apply the rules for the "in table body" insertion mode, the user agent must handle the token as follows:
    ...
    A start tag whose tag name is one of: "th", "td"
    Parse error.

    Clear the stack back to a table body context. (See below.)

    Insert an HTML element for a "tr" start tag token with no attributes, then switch the insertion mode to "in row".

    Reprocess the current token.

andy-h-chen:
@mikesamuel, can you please review?
[0001-For-th-and-td-add-tr-as-blockContainerParent.-When-i.txt](https://github.com/OWASP/java-html-sanitizer/files/371868/0001-For-th-and-td-add-tr-as-blockContainerParent.-When-i.txt)

mikesamuel:
Duplicate of #137

jmanico:
Backlog triage ahead of the next release.

Closing as a duplicate of #137, the table-formatting umbrella issue, as @mikesamuel noted in this thread. The test case here is worth keeping — please follow #137 for progress on table balancing, and re-open or comment there if this specific case is not covered.

PR #81, opened 2016-07-21
For 'th' and 'td', add 'tr' as blockContainerParent. When it has to c…

andy-h-chen:
…lose elements that cannot

contain 'td' or 'th', check if it could contain the blockContainerParent, insert blockContainerParent
if yes.
This is to address issue 77.


andy-h-chen:
This is for #77

andy-h-chen:
@mikesamuel, can you please review?

Issue #89, opened 2016-09-19
Allow hyphens in font family name.


pukomuko:
This bug was already open a few times:
https://github.com/OWASP/java-html-sanitizer/issues/27
https://github.com/OWASP/java-html-sanitizer/issues/10

We need to specify fonts with their styles:
https://coderwall.com/p/gdpyjq/how-to-list-all-available-fonts-in-xcode-by-family-name

What steps will reproduce the problem?

```java
String css = "font-family:'ChalkboardSE-Light', 'Arial'";
StylingPolicy stylingPolicy = new StylingPolicy(CssSchema.DEFAULT);
stylingPolicy.sanitizeCssProperties(css);
```

What is the expected output? What do you see instead?
Expected: `font-family: 'chalkboardse-light' , 'arial'`
Actual: `font-family: , 'arial'`

There are two separate problems here:

1. Font names with hyphens are removed.
2. Comma is in incorrect place resulting in invalid css.

mikesamuel:
PR was good after rerun.  Merged, added some tests.  Will release.

PR #90, opened 2016-09-19
allow hyphens in font family name

pukomuko:
this fixes bug #89


mikesamuel:
Thanks for the PR.  Sorry it's taken me so long to respond.

Does this allow font-family names like `---`?  If so, is that a valid family name?

pukomuko:
It does not allow for '---' family name. I cannot tell if it is valid name. Probably not.

mikesamuel:
Thanks for the patch.   I'm going to add some tests to make sure there's no CDO/CDC ambiguity ( `<!--`, `-->` ) but it'll be in the next release.

PR #97, opened 2016-10-14
Fix StringIndexOutOfBoundsException when sanitizing style="-: 0".

lquinn:


lquinn:
@mikesamuel, we see this quite frequently in the wild when sanitizing HTML email. The fix is pretty simple--could we pick it up for a future release?

mikesamuel:
Thanks for the patch.  I think this is obviated by https://github.com/OWASP/java-html-sanitizer/commit/0789289d5c5cddab361c17f1b90f38d14a2605c1 .  Sorry for not checking open PRs before fixing.

PR #108, opened 2017-03-16
Fix an out of bound error in CssSchema.java

jerryzz0:
Fix an index out of bound error.


mikesamuel:
Thanks for the patch, but I fixed this in https://github.com/OWASP/java-html-sanitizer/commit/0789289d5c5cddab361c17f1b90f38d14a2605c1 before seeing this PR.

PR #119, opened 2017-07-04
Added encodingPolicies Map to make text encoding configurable

yashkumar18:
Added a <Character,String> Map to store custom text encoding policies, to override default behaviour. In case, user wants to excludes encoding of selected characters or for all of them as per their requirement. The Map stores characters for which the user decides to disable encoding. If custom policies are not created while building Policy (i.e. the Map is empty or null) then the default behaviour is used (Encoding.REPLACEMENT).
This entity (encodingPolicies) is added as a member in HTMLPolicyBuilder and HTMLStreamRenderer. Required functions are overloaded to pass encodingPolicies as null if that parameter is missing. (For eg: when these functions are called without PolicyBuilder, as some of them are static functions)
null encodingPolicies results in default encoding behaviour.
None of the previous functionality will be affected since the functions that are required by this feature are overloaded to pass encodingPolicies as null.

### DIFF
--- a/src/main/java/org/owasp/html/Encoding.java
+++ b/src/main/java/org/owasp/html/Encoding.java
@@ -29,6 +29,7 @@
 package org.owasp.html;
 
 import java.io.IOException;
+import java.util.Map;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


mikesamuel:
I appreciate the thought and effort that went into this, but no.

I don't see that this addresses an actual problem.  As far as I can tell, this adds a configuration option for purely cosmetic reasons.

Adding configuration options
1.  requires testing more configurations reducing the amount of time maintainers have to rollout non-cosmetic changes,
2.  expands the attack surface without increasing the tested attack surface

Allowing policy authors to avoid encoding or mis-encoding meta-characters increases the potential that some policy authors will shoot themselves in the foot.

PR #126, opened 2017-10-12
Integrate url classifier

mikesamuel:
Integrated URL classifiers from github/OWASP/url-classifier
to make it easier to vet URLs without resorting to error-prone
regexs.

This is experimental.

This change deprecates the old way of white-listing protocols
in the scope of the whole sanitizer and makes available an
AttributePolicyBuilder match method that takes a classifier
so it is easy to scope a URL policy to a particular attribute.

Converting an attribute value to a UrlValue requires a UrlContext.
A UrlContext can encapsulate the base URL of the embedding document,
so should not be scoped to a policy, but rather to a sanitize call.
Therading the context through to Attribute and Element policies
requires an API change to {Attribute,Element}Policy.
This change deprecates those in favor of {Attribute,Element}Policy.V2
which are subinterfaces that take an extra Context object.


doug-macdonald17:
@mikesamuel - was this ever considered to be merged?

jmanico:
Closing this experimental branch to clear the queue. It has conflicted with `main` for years and the url-classifier integration was never finished. The branch is preserved on this repository as `integrate-url-classifier`, so it can be revived by anyone, including @mikesamuel, who wants to pick it back up. For @doug-macdonald17 and others who want richer URL vetting today, an `AttributePolicy` on the URL attribute (for example via `allowAttributes("href").matching(...)`) is the supported route.

PR #134, opened 2018-01-10
Make canContain return true for <table>, <thead>, <tbody>, <tfoot> an…

zhajiang:
Make canContain return true for `<table>, <thead>, <tbody>, <tfoot> and <tr>` elements,
so that it will not close the open table elements and break table layout.

### DIFF
--- a/src/test/java/org/owasp/html/SanitizersTest.java
+++ b/src/test/java/org/owasp/html/SanitizersTest.java
@@ -313,7 +313,7 @@ public class SanitizersTest extends TestCase {
       .and(Sanitizers.STYLES)
       .and(Sanitizers.IMAGES)
       .and(Sanitizers.TABLES);
-    assertEquals("<table></table>Hallo\r\n\nEnde\n\r", pf.sanitize(input));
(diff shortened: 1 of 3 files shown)


waiyan1997:
95 &nbsp;9597******30

src/test/java/org/owasp/html/SanitizersTest.java
```diff
@@ -313,7 +313,7 @@ public static final void testScriptInTable() {
       .and(Sanitizers.STYLES)
       .and(Sanitizers.IMAGES)
       .and(Sanitizers.TABLES);
-    assertEquals("<table></table>Hallo\r\n\nEnde\n\r", pf.sanitize(input));
+    assertEquals("<table>Hallo\r\n\nEnde\n\r</table>", pf.sanitize(input));
```
mikesamuel: Why is this a better behavior?

mikesamuel:
Thanks for the patch.

I think https://github.com/OWASP/java-html-sanitizer/commit/66794b6f3092f9f4024aa82ed75af93f8954f5d2 addresses this.

PR #161, opened 2018-12-20
Allow import of OWASP AntiSamy XML policy files

efge:
Please review this to allow simple parsing of AntiSamy files.


efge:
Travis build failed due a timeout fetching some artifacts, I don't know how to relaunch it.

mikesamuel:
It looks like there's 2 points of intersection.
*  Adding rules to an HtmlPolicyBuilder from an AntiSamyPolicy
*  Checking attribute values using an AntiSamyAttribute.

Instead of having HtmlPolicyBuilder fetch and inspect an AntiSamyPolicy, how about the AntiSamiPolicy modifies an HtmlPolicyBuilder?

Then instead of including in HtmlPolicyBuilder the matcher for an AntiSamyAttribute, that could live as a custom matcher in the method that adds rules to a policy builder.

That way, this code could be entirely separate from the main sanitizer and be published as an independent maven package.

mikesamuel:
Once we figure out how to separate them, it'd be nice if the test suite made sure they integrate nicely.
Maybe having the antisamy code as a sibling package of the sanitizer would make it easy to ensure that.

efge:
I had chosen to modify as little as possible the original AntiSamy code given that's it's mostly dead, but I can certainly switch it around to what you suggest and not modify base `HtmlPolicyBuilder` code, to make it not depend on AntiSamy.

I'll make it a sibling package then. Will work on this in a few days.

kwin:
@efge Are you still working on this?

efge:
@kwin sorry no I haven't been able to find the time to do this. I don't foresee any free time in the coming weeks either.

Issue #151, opened 2018-07-17
skipRelsOnLinks causes StringIndexOutOfBoundsException


hikoma:
Version: 20180219.1

Enabling skipRelsOnLinks causes StringIndexOutOfBoundsException when an `a` tag having `target` without `rel` exists.

```
import org.junit.jupiter.api.Test;
import org.owasp.html.HtmlPolicyBuilder;

class SanitizeTest {
    @Test
    public void test() {
        new HtmlPolicyBuilder()
                .allowElements("a")
                .allowAttributes("href", "target").onElements("a")
                .skipRelsOnLinks("noopener", "noreferrer")
                .toFactory()
                .sanitize("<a href='' target='_blank'>Hello</a>");
    }
}
```
```
java.lang.StringIndexOutOfBoundsException: String index out of range: -1

	at java.lang.AbstractStringBuilder.substring(AbstractStringBuilder.java:935)
	at java.lang.StringBuilder.substring(StringBuilder.java:76)
	at org.owasp.html.HtmlPolicyBuilder$RelsOnLinksPolicy.apply(HtmlPolicyBuilder.java:1007)
	at org.owasp.html.ElementAndAttributePolicyBasedSanitizerPolicy.applyPolicies(ElementAndAttributePolicyBasedSanitizerPolicy.java:145)
	at org.owasp.html.ElementAndAttributePolicyBasedSanitizerPolicy.openTag(ElementAndAttributePolicyBasedSanitizerPolicy.java:105)
	at org.owasp.html.TagBalancingHtmlStreamEventReceiver.openTag(TagBalancingHtmlStreamEventReceiver.java:114)
	at org.owasp.html.HtmlSanitizer.sanitize(HtmlSanitizer.java:193)
	at org.owasp.html.PolicyFactory.sanitize(PolicyFactory.java:127)
	at org.owasp.html.PolicyFactory.sanitize(PolicyFactory.java:107)
	at SanitizeTest.test(SanitizeTest.java:12)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #163, opened 2019-01-31
Prevent StringIndexOutOfBoundsException on empty default rel set

kyle-simmons:
When using the `HtmlPolicyBuilder`, if both `noopener` and  `noreferrer` are supplied to `skipRelsOnLinks`, then a `StringIndexOutOfBoundsException` will get thrown when applying sanitization on elements that have both an `href` and `target`. This PR addresses that issue by only attempting to apply rels to the element if we actually have a non-zero set of rels to apply.

Addresses https://github.com/OWASP/java-html-sanitizer/issues/151

### DIFF
--- a/src/main/java/org/owasp/html/HtmlPolicyBuilder.java
+++ b/src/main/java/org/owasp/html/HtmlPolicyBuilder.java
@@ -987,7 +987,7 @@ public class HtmlPolicyBuilder {
       if (indexOfAttributeValue("href", attrs) >= 0) {
         // It's a link.
         boolean hasTarget = indexOfAttributeValue("target", attrs) >= 0;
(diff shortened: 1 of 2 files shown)


kyle-simmons:
@mikesamuel Could you review this PR and provide feedback on if this is behavior we want to allow? I think regardless a `StringIndexOutOfBoundsException` is not ideal.

mikesamuel:
Sorry for the delay.  Thanks for the testcase.  I think there's a more minimal fix at L1008.

mikesamuel:
Folded your test and a tweaked fix at https://github.com/OWASP/java-html-sanitizer/commit/83c4ce38014d42709c8262804bba8254dfa2a30c

PR #174, opened 2019-05-24
Add missing mappings for named HTML entities

jacobrshields:
This adds missing mappings for named HTML entities.

Currently, certain entities are not recognized, so for instance `&DoubleUpArrow;` gets "sanitized" to `&amp;DoubleUpArrow;` (incorrect) instead of `⇑` (correct).

I used a piece of JavaScript to generate the Java mappings based on the data at https://dev.w3.org/html5/html-author/charref.

JavaScript used to scrape the entities:

```js
// Execute in browser at https://dev.w3.org/html5/html-author/charref
(function () {
  let lastCategory = null;

  for (tr of document.querySelectorAll('tr')) {
    const category = tr.getAttribute('data-block');
    if (category !== lastCategory) {
      console.log(`\n      // ${category}`);
    }
    lastCategory = category;

    const description = tr.querySelector('td.desc').innerText;
    const names = tr.querySelector('td.named code').innerText.split(' ');

    const rawHexStr = tr.querySelector('td.hex code').innerHTML; // '&#x1D56B;'
    const unicodeStr = rawHexStr.substring(rawHexStr.indexOf('x') + 1, rawHexStr.length - 1); // '1D56B'
    const unicodeValue = parseInt(unicodeStr, 16); // 0x1D56B

    let upperUnicodeValue;
    let lowerUnicodeValue;
    if (unicodeValue < 0x10000) {
      upperUnicodeValue = 0;
      lowerUnicodeValue = unicodeValue;
    } else {
      const difference = (unicodeValue - 0x10000);
      upperUnicodeValue = 0xD800 | (difference >> 10); // 0xd835
      lowerUnicodeValue = 0xDC00 | (difference & 0x3FF); // 0xdd6b
    }

    const upperUnicodeStr = upperUnicodeValue.toString(16).padStart(4, '0'); // 'd835'
    const lowerUnicodeStr = lowerUnicodeValue.toString(16).padStart(4, '0'); // 'dd6b'

    const javaValue = (upperUnicodeStr === '0000'
      ? `Integer.valueOf('\\u${lowerUnicodeStr}')`
      : `Character.toCodePoint('\\u${upperUnicodeStr}', '\\u${lowerUnicodeStr}')`)
      .replace('\\u000a', '\\n')
      .replace('\\u0027', '\\\'')
      .replace('\\u005c', '\\\\');

    for (const name of names) {
      console.log(`      builder.put("${name.substring(1, name.length - 1)}", ${javaValue}); // ${description}`);
    }
  }
})();
```


src/main/java/org/owasp/html/HtmlEntities.java
```diff
@@ -217,292 +231,2104 @@ public static long decodeEntityAt(String html, int offset, int limit) {
 //  }
 
   /** A trie that maps entity names to codepoints. */
-  public static final Trie ENTITY_TRIE = new Trie(
-      ImmutableMap.<String, Integer>builder()
+  public static final Trie ENTITY_TRIE = new Trie(getEntityNameToCodePointMap());
+
+  private static Map<String, Integer> getEntityNameToCodePointMap() {
```
binaryword: I had to refactor this slightly because the thousands of chained `.put()` calls caused the Java compiler to run out of resources during compilation, due to it tracking the entire stack.

src/main/java/org/owasp/html/HtmlEntities.java
```diff
@@ -28,13 +28,27 @@
...
+  private static final int LONGEST_ENTITY_NAME;
+
+  static {
+    int longestEntityName = 0;
+    for (String entityName : getEntityNameToCodePointMap().keySet()) {
+      if (entityName.length() > longestEntityName) {
+        longestEntityName = entityName.length();
+      }
+    }
+    LONGEST_ENTITY_NAME = longestEntityName;
```
binaryword: Since I had to factor out the generation of the map to a separate method anyway, I decided to dynamically calculate this value. However if you don't like the static initializer we could hardcode it as well. The longest entity name is CounterClockwiseContourIntegral, which is 31 characters.

src/main/java/org/owasp/html/HtmlEntities.java
```diff
@@ -28,13 +28,2127 @@
...
  */
 final class HtmlEntities {
 
+  /** A trie that maps entity names to codepoints. */
+  public static final Trie ENTITY_TRIE;
+
+  private static final int LONGEST_ENTITY_NAME;
+
+  static {
+    final ImmutableMap.Builder<String, Integer> builder = ImmutableMap.builder();
```
binaryword: I had to refactor the map building to not used chained calls, because having thousands of chained `.put()` calls caused the Java compiler to hit a stack overflow during compilation unless it was given a larger stack size via `-Xss`.

src/main/java/org/owasp/html/HtmlEntities.java
```diff
@@ -28,13 +28,2127 @@
...
 
 /**
  * Utilities for decoding HTML entities, e.g., {@code &amp;}.
  */
 final class HtmlEntities {
 
+  /** A trie that maps entity names to codepoints. */
+  public static final Trie ENTITY_TRIE;
+
+  private static final int LONGEST_ENTITY_NAME;
```
binaryword: This comes out to be `31` (for `"CounterClockwiseContourIntegral"`).

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -37,7 +37,7 @@
 
   /**
    * Decodes HTML entities to produce a string containing only valid
-   * Unicode scalar values.
+   * Unicode code points.
```
binaryword: Updated since this works on surrogate pairs which are not scalar values.

mikesamuel: The difference between scalar values and code points are orphaned surrogates so I don't see what surrogate pairs have to do with this.

src/main/java/org/owasp/html/HtmlEntities.java
mikesamuel: Fair enough.  What's actually changed here?

src/main/java/org/owasp/html/HtmlEntities.java
```diff
@@ -28,13 +28,2129 @@
...
+
+  static {
+    final ImmutableMap.Builder<String, Integer> builder = ImmutableMap.builder();
+
+    // Source data: https://dev.w3.org/html5/html-author/charref
```
mikesamuel: Why prefer this to the relevant section of the HTML spec? https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references

src/main/java/org/owasp/html/HtmlEntities.java
```diff
@@ -28,13 +28,2129 @@
...
+    builder.put("downdownarrows", Integer.valueOf('\u21ca')); // DOWNWARDS PAIRED ARROWS
+    builder.put("lrhar", Integer.valueOf('\u21cb')); // LEFTWARDS HARPOON OVER RIGHTWARDS HARPOON
+    builder.put("ReverseEquilibrium", Integer.valueOf('\u21cb')); // LEFTWARDS HARPOON OVER RIGHTWARDS HARPOON
+    builder.put("leftrightharpoons", Integer.valueOf('\u21cb')); // LEFTWARDS HARPOON OVER RIGHTWARDS HARPOON
+    builder.put("rlhar", Integer.valueOf('\u21cc')); // RIGHTWARDS HARPOON OVER LEFTWARDS HARPOON
+    builder.put("rightleftharpoons", Integer.valueOf('\u21cc')); // RIGHTWARDS HARPOON OVER LEFTWARDS HARPOON
+    builder.put("Equilibrium", Integer.valueOf('\u21cc')); // RIGHTWARDS HARPOON OVER LEFTWARDS HARPOON
+    builder.put("nlArr", Integer.valueOf('\u21cd')); // LEFTWARDS DOUBLE ARROW WITH STROKE
+    builder.put("nLeftarrow", Integer.valueOf('\u21cd')); // LEFTWARDS DOUBLE ARROW WITH STROKE
+    builder.put("nhArr", Int
```
mikesamuel: Why not just use Integer.valueOf with the codepoint instead of doing an unnecessary conversion from surrogates?

Fyi, https://html.spec.whatwg.org/entities.json includes codepoints as well as the string form.

src/test/java/org/owasp/html/EncodingTest.java
```diff
@@ -188,6 +188,12 @@ public static final void testDecodeHtml() {
     assertEquals(
         "\u03b1",
         Encoding.decodeHtml("&alpha;"));
+    assertEquals(
+        "\ud835\udc9c",  // U+1D49C requires a surrogate pair in UTF-16.
+        Encoding.decodeHtml("&Ascr;"));
+    assertEquals(
+        "\u2233",  // HTML entity with the longest name.
+        Encoding.decodeHtml("&CounterClockwiseContourIntegral;"));
```
mikesamuel: omfg

src/main/java/org/owasp/html/HtmlEntities.java
jacobrshields: The structural change is that the old code was of the form `builder.put().put().put().build()` while the new code is of the form `builder.put(); builder.put(); builder.put(); builder.build()` (+ a static initialization block). The difference being the second form evidently uses less stack space during compilation. Functionally the change is the addition of all the entity mappings.

src/main/java/org/owasp/html/Encoding.java
jacobrshields: :+1: I misunderstood. Will revert.

src/test/java/org/owasp/html/EncodingTest.java
jacobrshields: Haha, yep...

src/main/java/org/owasp/html/HtmlEntities.java
jacobrshields: No reason. It was the source I found. I will update to use the linked spec. :+1:

src/main/java/org/owasp/html/HtmlEntities.java
jacobrshields: Because Java characters are 16 bits and this codepoint requires the use of a surrogate to express in UTF-16, i.e. it cannot be expressed as a single Java character. Therefore we would have to use `Integer.valueOf(String)` which has a different behavior. Unless I'm misunderstanding your question.

src/main/java/org/owasp/html/Encoding.java
jacobrshields: Reverted

src/main/java/org/owasp/html/HtmlEntities.java
jacobrshields: Updated. Evidently some named references refer to 2 code points. Since all such references only refer to BMP code points, we can still pack those 2 code points into a 32-bit integer and not have to change the signature of `decodeEntityAt`, but I had to add some extra bit shifting.

jacobrshields:
Updated script to generate mappings, using the data downloaded from https://html.spec.whatwg.org/entities.json:

```js
#!/usr/bin/env node
'use strict';

const fs = require('fs');

const ENTITIES_JSON_PATH = '/Users/jshields/entities.json';

const entities = JSON.parse(fs.readFileSync(ENTITIES_JSON_PATH));

for (const entityKey of Object.keys(entities)) {
  if (entityKey[entityKey.length - 1] !== ';') {
    continue;
  }

  const entityName = entityKey.substring(1, entityKey.length - 1);
  const characters = entities[entityKey].characters;

  if (characters.length === 1) {
    console.log(`builder.put("${entityName}", ((int) ${getJavaCharacterLiteral(characters[0])}) << 16);`);
  } else if (characters.length === 2) {
    console.log(`builder.put("${entityName}", (((int) ${getJavaCharacterLiteral(characters[0])}) << 16) | ${getJavaCharacterLiteral(characters[1])});`);
  } else {
    throw new Error(`Unexpected number of characters for ${entityKey}`);
  }
}

function getJavaCharacterLiteral(character) {
  return `'\\u${character.charCodeAt(0).toString(16).padStart(4, '0')}'`
    .replace('\\u000a', '\\n')
    .replace('\\u0027', '\\\'')
    .replace('\\u005c', '\\\\');
}
```

src/main/java/org/owasp/html/HtmlEntities.java
mikesamuel: Why does it matter what Java characters are?  I just need an input I can use with `StringBuilder.appendCodepoint`.

Actually, I suppose for `"&NotEqualTilde;": { "codepoints": [8770, 824], "characters": "\u2242\u0338" },`  I really need a string that I can use with `StringBuilder.append`.

Perhaps this map should change to just store the replacement string which is just the stringified version of the chars but with backslash and double quotes escaped Java style.

jacobrshields: You're right, it doesn't matter. It could work either way and the output will be the same. This is executed once during class load and it's not clear that the performance will be noticeably different either way. Personally I think `\u2242` is more readable / traceable for the developer than `8770` but I think this is just an issue of code style. If you prefer it another way I can update it.

As you notice things are a bit tricker with references that refer to multiple code points. We could update `decodeEntityAt` to return a variable-length array. It turns out we can also stuff the code points into 32 bits, which is what I have now. The speculative idea being that it gives better performance because we don't have to allocate an array on the heap. But is it a noticeable difference? I don't know. We can do it either way.

mikesamuel:
Thanks for this.
What if we just stored strings like

```js
#!/usr/bin/env node
'use strict';

const url = require('url');
const fetch = require('node-fetch');

const ENTITIES_JSON_URL = 'https://html.spec.whatwg.org/entities.json';

(async () => {
  const response = await fetch(ENTITIES_JSON_URL, { method: 'GET' });
  if (response.status !== 200) { throw new Error(); }
  const entities = JSON.parse(await response.text());

  for (const entityKey of Object.keys(entities)) {
    if (entityKey[entityKey.length - 1] !== ';') {
// Are you sure this is right?
// I think an entityKey of "&lt" means that "&lt" need not end with a semicolon.
// IIRC, this is what allows recognizing that in <a href="?a=b&ampx=0&copy=1">
// the "&amp" means "&" but "&copy" is not a copyright symbol.
      continue;
    }

    const entityName = entityKey.substring(1, entityKey.length - 1);
    const { characters } = entities[entityKey];

    console.log(`    builder.put("${entityName}", ${ javaString(characters) });`);
  }
})().then(() => {}, (e) => console.error(e));

const specialCharacters = new Map();
specialCharacters.set('\\', String.raw`\\`);
specialCharacters.set('"', String.raw`\"`);

function javaString(s) {
  const javaChars = '"' +
      s.replace(
        /[\s\S]/g,
        (c) => specialCharacters.has(c) ? specialCharacters.get(c) : '\\u' + hex4(c.charCodeAt(0)))
      + '"';
  if (s !== JSON.parse(javaChars)) { throw new Error(s); }
  return javaChars;
}

function hex4(n) {
  const hex = n.toString(16);
  return '0000'.substring(hex.length) + hex;
}
```

and then I think Encoding.decodeHtml s the only code that calls `decodeEntityAt`, so `decodeEntityAt` could change to `appendDecodedEntity`, take a StringBuilder as input, and return the `int end`.

https://github.com/OWASP/java-html-sanitizer/blob/7cdb5eb73b3d32d202d6a0b6ffc28492ae9b3099/src/main/java/org/owasp/html/Encoding.java#L57-L60

jacobrshields:
Good idea. I like the idea of `appendDecodedEntity`.

Regarding references that don't end with `;`: I think this is a separate pre-existing issue that we may want to handle under the scope of a different PR. Yes it appears that `&amp`, `&lt`, etc. are legal character references and browsers do display them as expected. But the current code first parses the string between `&` and `;` and then checks if it's a numeric entity or a named entity.

Actually the spec is a little confusing in that [in one place](https://html.spec.whatwg.org/multipage/syntax.html#character-references) it says that **must** end with `;`:

>  The name must be one that is terminated by a U+003B SEMICOLON character (;).

But the character reference sheet clearly shows that not all of them do.

Are you OK de-scoping that particular concern to a different PR?

mikesamuel:
> Are you OK de-scoping that particular concern to a different PR?

Yes.

jacobrshields:
OK, sounds good.

I updated to use your suggestion of `appendDecodedEntity`. Also I had to add 1 more special character mapping to your JavaScript for the newline character.

Also curious what your thoughts are on this comment: https://github.com/OWASP/java-html-sanitizer/pull/174#pullrequestreview-241990112

jacobrshields:
@mikesamuel When would you be able to take another look at this PR for approval? Would love to get this merged soon if possible for a project I am working on.

scottastrophic:
>I believe named entities are case-sensitive.

They are indeed. You can see why with accented characters: `&Eacute;` is É while `&eacute;` is é.

mikesamuel:
Sorry for the delay.  Was slammed with deadlines.  Taking another look.

src/main/java/org/owasp/html/HtmlEntities.java
```diff
@@ -182,13 +2336,16 @@ public static long decodeEntityAt(String html, int offset, int limit) {
...
       }
     }
     if (codepoint < 0) {
-      return ((offset + 1L) << 32) | '&';
+      sb.append('&');
+      return offset + 1;
     } else {
-      return (((long) tail) << 32) | codepoint;
+      sb.appendCodePoint(codepoint);
+      return tail;
```
mikesamuel: Great.  All this looks semantics preserving modulo the contents of the Trie.

src/main/java/org/owasp/html/Trie.java
```diff
@@ -41,20 +41,20 @@
...
  * @author Mike Samuel (mikesamuel@gmail.com)
  */
-final class Trie {
+final class Trie<T> {
   private final char[] childMap;
-  private final Trie[] children;
+  private final Trie<T>[] children;
   private final boolean terminal;
-  private final int value;
+  private final T value;
```
mikesamuel: Please add `@Nullable`.

src/main/java/org/owasp/html/Trie.java
```diff
@@ -103,26 +104,26 @@ private Trie(
...
     childMap[childIndex] = lastCh;
-    children[childIndex++] = new Trie(elements, depth + 1, childStart, end);
+    children[childIndex++] = new Trie<T>(elements, depth + 1, childStart, end);
   }
 
   /** Does this node correspond to a complete string in the input set. */
   public boolean isTerminal() { return terminal; }
 
-  public int getValue() { return value; }
+  public T getValue() { return value; }
```
mikesamuel: Please add `@Nullable`

src/test/java/org/owasp/html/EncodingTest.java
```diff
@@ -188,6 +188,15 @@ public static final void testDecodeHtml() {
...
         Encoding.decodeHtml("&alpha;"));
+    assertEquals(
+        "\ud835\udc9c",  // U+1D49C requires a surrogate pair in UTF-16.
+        Encoding.decodeHtml("&Ascr;"));
+    assertEquals(
+        "fj",  // &fjlig; refers to 2 characters.
+        Encoding.decodeHtml("&fjlig;"));
+    assertEquals(
+        "\u2233",  // HTML entity with the longest name.
+        Encoding.decodeHtml("&CounterClockwiseContourIntegral;"));
```
mikesamuel: Cool.  Maybe suss out IndexOutOfBounds via

```js
   assertEquals(
        // HTML entity with the longest name missing terminator.
        "&CounterClockwiseContourIntegral",  
        Encoding.decodeHtml("&CounterClockwiseContourIntegral"));
```

src/main/java/org/owasp/html/Trie.java
jacobrshields: Done

src/main/java/org/owasp/html/Trie.java
jacobrshields: Done

src/test/java/org/owasp/html/EncodingTest.java
jacobrshields: Sorry, I'm not sure I follow. Can you explain the case you want to protect against here?

Incidentally that input string does get decoded to `∳` because there is some logic for detecting "broken entities" that end with a non-alphanumeric character or end of string.

mikesamuel: Since that's the longest, it's the most likely culprit in any boundary bug that leads to attempts to read past the end of input looking for a semicolon or another character in a character reference name.

> Incidentally that input string does get decoded to ∳

Changing the expected result to `"\u2233"` seems fine.

jacobrshields: I see, got it. Added.

mikesamuel:
Thanks much for doing this.

I'll push a version with this.

jacobrshields:
Thank you Mike, I appreciate it!

mikesamuel:
https://github.com/OWASP/java-html-sanitizer/releases/tag/release-20190610.1 should be available at https://search.maven.org/artifact/com.googlecode.owasp-java-html-sanitizer/aggregate/20190610.1/pom shortly.

Issue #183, opened 2019-09-04
style attribute "margin" property is missing fourth element after sanitizing


janis-github:
The margin property is missing left alignment value after sanitisation.

Here is the code:
```java
     private static final CssSchema MY_CSSSCHEMA = CssSchema.withProperties(ImmutableSet.of(
          "margin",
          "list-style-type",
          "vertical-align"
  ));

   PolicyFactory policy = new HtmlPolicyBuilder()
        .allowAttributes("class")
        .onElements("li")
        .allowAttributes("checked", "type")
        .onElements("input")
        .allowElements("i", "u", "b", "p", "pre", "blockquote", "tt", "em", "ul", "ol", "br", "li", "sub", "sup", "div",
            "table", "img", "a", "tbody", "tr", "td", "strong", "span", "h1", "h2", "h3", "code", "md", "video",
            "source", "th", "thead", "h4", "h5", "h6", "h7", "h8", "h9", "h10", "h11", "hr")
        .allowStyling(MY_CSSSCHEMA)
        .toFactory();
    String postSanitize = policy.sanitize(text).trim();
```

This is my input string:

```
<ul>
<li class="task-list-item" style="list-style-type: none;"><input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;" checked> This is a complete item</li>
<li class="task-list-item" style="list-style-type: none;"><input type="checkbox" disabled style="margin: 0px 0.35em 0.25em -1.6em; vertical-align: middle;"> This is an incomplete item</li>
</ul>
```

Here is the output:

```
<ul><li class="task-list-item"><input type="checkbox" style="margin:0px 0.35em 0.25em" checked="checked" disabled="1" /> This is a complete item</li><li class="task-list-item"><input type="checkbox" style="margin:0px 0.35em 0.25em" disabled="1" /> This is an incomplete item</li></ul>
```

Note that 
```
style="margin:0px 0.35em 0.25em"
```
but it should be:
```
style="margin: 0px 0.35em 0.25em -1.6em"
```

janis-github:
Turns out that this is about "margin" token value being negative. Created PR184 for fixing that. Closing the issue in favour of PR.

PR #184, opened 2019-09-05
CssSchema: Allow negative token values for margin property

janis-github:
This fixes the https://github.com/OWASP/java-html-sanitizer/issues/183

### DIFF
--- a/src/main/java/org/owasp/html/CssSchema.java
+++ b/src/main/java/org/owasp/html/CssSchema.java
@@ -654,7 +654,7 @@ public final class CssSchema {
     Property listStyleType = new Property(
         0, union(listStyleLiterals0, listStyleLiterals2), zeroFns);
     builder.put("list-style-type", listStyleType);
-    Property margin = new Property(1, bottomLiterals0, zeroFns);
+    Property margin = new Property(5, bottomLiterals0, zeroFns);
     builder.put("margin", margin);
     Property maxHeight = new Property(1, maxHeightLiterals0, zeroFns);
     builder.put("max-height", maxHeight);
--- a/src/main/java/org/owasp/html/CssSchema.java
+++ b/src/main/java/org/owasp/html/CssSchema.java
@@ -633,7 +633,7 @@ public final class CssSchema {
     Property fontWeight = new Property(
         0, union(fontLiterals0, fontStyleLiterals0), zeroFns);
     builder.put("font-weight", fontWeight);
-    Property height = new Property(5, bottomLiterals0, zeroFns);
+    Property height = new Property(1, bottomLiterals0, zeroFns);
     builder.put("height", height);
     Property letterSpacing = new Property(5, fontStyleLiterals0, zeroFns);
     builder.put("letter-spacing", letterSpacing);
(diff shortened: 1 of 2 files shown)


janis-github:
@mikesamuel, 

How could my fix possibly break the build system?
And reduce coverage in completely different/unmodified file?

Thanks,
Janis.

mikesamuel:
> How could my fix possibly break the build system?
> And reduce coverage in completely different/unmodified file?

@janis-github The baseline run in Travis isn't super old, but sometimes this happens when Travis has changed something like the set of supported JDKs, or a new version of the maven test coverage plugin does things subtly differently.  If we resolve other issues, I'm happy to pull and merge manually.

mikesamuel:
Thanks for the patch.

There used to be a problem with negative margins allowing HTML like the below to escape visual containment mechanisms like CSS `overflow`.

```html
<!doctype html>
<html>
  <meta charset="utf-8" />
<body>

<!--  trusted content -->
<form style="float: right; width: 10em" id="sidebar">
  <label for="account">Email:</label> <input type="text" name="account">
  <br><label for="pwd">Password:</label> <input type="password" name="pwd"></p>
</form>
<div style="width: 30em">
  <p>Lorem ipsum sic dolor amet
  <p>Lorem ipsum sic dolor amet
  <p>Lorem ipsum sic dolor amet
  <p>Lorem ipsum sic dolor amet

  <h2>Comments</h2>
  <div style="overflow: auto">
<!-- /trusted content -->


<!--  sanitized, third-party content -->
    <p>Comment content</p>

    <p style="margin: -8em 0 8em 25em">Contact user support +1-800-PHISHER</p>
<!-- /sanitized, third-party content -->

  </div>
</html>
```

It used to be the case, that certain browsers (sorry I forget which) used to render that HTML thus:

----

Screen Shot 2019-09-05 at 12 47 10 PM

----

Note that the "Contact user support +1-800-PHISHER" text appears below the password input where it seems to be part of the high-reputation login form, instead of being constrained to the rectangle for low-reputation content.

Before allowing negative margins, I'd want to double check that browsers no longer have that bug.

Would replacing negative values with `0` instead of dropping them when non-negative values are allowed provide a short-term improvement?

janis-github:
I see your point. I was thinking about that and similar behaviour can be achieved with large positive margins going out of the box, too. (Do not have sample for that, though) There is no solution based only on margin values. So, normalising negatives to 0 (or any other number for that matter) does not solve the problem. In my specific case which triggered this issue setting right margin to -1.6em did actually make the output prettier.
To be 100% sure sanitiser would need to render the html and "see" if there are any suspicious overlaps, etc... But that would be well beyond the the scope of this lib, wouldn't it?
Maybe normalise to some sane value like up to -5em/-50px is ok, but anything less than that is bad? And make those configurable (separate for vertical/horizontal, top/margin-top, left/right, specific for each element, etc...). I do not know if that's worth it, really.

mikesamuel:
@janis-github, Sorry for the delay.  The problem was that visual cropping was bypassed when margins were negative.  Not that you can move things around.  I'm not aware of any numeric overflow problems that bypassed visual cropping with large margins.

janis-github:
Ok, resetting negative margins to 0 probably is the best solution then. Sorry, I'll not have a PR as it involves going significantly deeper in your code as I'd like, currently. I.e. resetting token values to zero instead of just ignoring.

jmanico:
Closing this one out. As @mikesamuel explained above, negative margins were deliberately excluded because they let sanitized content escape a visual container that uses `overflow` clipping, which is a real attack surface for click-jacking and UI redress inside embedded content. @janis-github agreed in the last comment that clamping negatives to zero would be the better approach, and that is effectively what the current behaviour does by dropping them. If someone wants to pursue the clamp-to-zero variant, please open a fresh PR against `main` and reference this one. Thanks for the patch and the discussion.

PR #185, opened 2019-09-17
Add OSGI bundle packaging

benapple:
In its current state, this project does not work out-of-the-box within an OSGi environment. Adding the [Maven bundle plugin](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html) and changing the packaging to bundle allows the project to be used more seamlessly. Without the resulting manifest files of the plugin, developers either need to embed this dependency in all bundles that use the project or need to separately wrap the java-html-sanitizer jar as a bundle using the standalone BND tool.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,9 +60,23 @@
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
-        <extensions>true</extensions>
-        <version>3.5.0</version>
       </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-verifier-plugin</artifactId>
+          <configuration>
+            <verificationFile>src/test/resources/osgi-integration-verification.xml</verificationFile>
+          </configuration>
+          <executions>
+            <execution>
+              <id>main</id>
+              <phase>verify</phase>
+              <goals>
+                <goal>verify</goal>
+              </goals>
+            </execution>
+          </executions>
+        </plugin>
     </plugins>
   </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -1,7 +1,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <artifactId>owasp-java-html-sanitizer</artifactId>
-  <packaging>jar</packaging>
+  <packaging>bundle</packaging>
   <parent>
     <relativePath>parent</relativePath>
     <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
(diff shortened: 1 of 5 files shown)


mikesamuel:
I'm horribly ignorant of OSGI bundling.  How does changing the packaging type affect non-OSGI clients?

Is there a standard way to test, pre-commit, that OSGI integration works?

mikesamuel:
Is ["Adding OSGi metadata to existing projects without changing the packaging type"](https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#adding-osgi-metadata-to-existing-projects-without-changing-the-packaging-type) insufficient?

benapple:
A bundle is a jar with some extra OSGi bits within the manifest that non-OSGi clients can safely ignore. For example, Guava uses bundle packaging without the java-html-sanitizer project needing to do anything special to use it. With that said, both the bundle packaging and the jar packaging with extra manifest-generating steps work equally well.

OSGi integration is not something I have done within a unit test, though I am taking a look at [osgi-mock](https://sling.apache.org/documentation/development/osgi-mock.html) to see what is possible. Ideally, the test would ensure the Maven plugin generates a usable manifest.

pom.xml
```diff
@@ -1,7 +1,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <artifactId>owasp-java-html-sanitizer</artifactId>
-  <packaging>jar</packaging>
+  <packaging>bundle</packaging>
```
mikesamuel: It'd be nice if this also happened to html-types/pom.xml and aggregate/pom.xml.

mikesamuel:
>  Ideally, the test would ensure the Maven plugin generates a usable manifest.

I wonder whether a shell test that looks at the output of

```sh
unzip -p target/owasp-*-SNAPSHOT.jar META-INF/MANIFEST.MF
```

would suffice.

pom.xml
```diff
@@ -57,6 +57,12 @@
           <repoToken>TENYw0IKRulLOXgkyzjGqNi0hQyhBiSiU</repoToken>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
```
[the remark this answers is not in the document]
benapple: I removed this `<extension>true</extensions>` as it wasn't adding anything necessary. I have moved the version into the parent though.

pom.xml
benapple: I've also added the bundle packaging to the html-types/pom.xml. The aggregate/pom.xml doesn't need updated as it will still use both the root pom and html-types/pom to build bundles.

benapple:
There turned out to be a somewhat handy maven plugin for verifying file contents from a build, [the Maven verifier plugin](https://maven.apache.org/plugins/maven-verifier-plugin/index.html). I've added just a simple check to ensure the main package of each module is specified in the Export-Package section of the manifest.

mikesamuel:
Thanks.  I'll pull this locally and see how it works.

mikesamuel:
Thanks alot for explaining bundling and for the PR.  Merged.

I plan to clean up some `mvn` warnings around missing versions.  Then I'll push a new version that is properly bundled.  I'll post here with a link to the new version when that's done.

mikesamuel:
Present in https://search.maven.org/artifact/com.googlecode.owasp-java-html-sanitizer/owasp-java-html-sanitizer/20191001.1/bundle

IIUC, the `/bundle` at the end of that URL reflects the change.

benapple:
Awesome! I've pulled the 20191001.1 version and confirmed my project activates the bundle and that my local services can reference `org.owasp.html.*` classes. Thanks for merging this!

Issue #193, opened 2020-02-16
decoded html entities in href


fr3akX:
href attribute value has decoded html entities in output.

Example
```java
        String input = "" +
                "<a href=\"/index.php?action=1&order_id=1\">order</a>";
        System.out.println(Sanitizers.LINKS.sanitize(input));
```

output
```
<a href="/index.php?action&#61;1ℴ_id&#61;1" rel="nofollow">order</a>
```

`&order` is decoded as `ℴ`, which renders invalid link

yangbongsoo:
@fr3akX Hello. I think `-` and `_` char may be treated in plain letters. What do you think?

PR #198, opened 2020-05-20
'-' and '_' may be treated in plain letters.

yangbongsoo:
#193 becaue of `_` char, `order` string  find it in ENTITY_TRIE and replace `ℴ` 
I think `-` and `_` may be treated in plain letters.
Please tell me if you think I'm wrong or if I lack a test.

### DIFF
--- a/src/main/java/org/owasp/html/HtmlEntities.java
+++ b/src/main/java/org/owasp/html/HtmlEntities.java
@@ -2232,6 +2232,7 @@ final class HtmlEntities {
         case 'y': case 'z':
         case '0': case '1': case '2': case '3': case '4': case '5':
         case '6': case '7': case '8': case '9':
+        case '-': case '_':
           break;
         case '=':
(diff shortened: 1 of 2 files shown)


mikesamuel:
Hmm.  Since this was initially written, https://html.spec.whatwg.org/#named-character-references clarified which named character references are allowed without a trailing semicolon.

I believe, according to html.spec.whatwg, that `&curren` is allowed without a semicolon because there are two entries in that table, one of which does not have a trailing `;`:

> | Name | Character | Glyph
> | -- | -- | --
> | curren; | U+000A4 | ¤
> | curren | U+000A4 | ¤

Maybe we should just derive a list of HTML entities that are allowed without semicolons instead of looking for extra letters.

mikesamuel:
Does https://github.com/OWASP/java-html-sanitizer/pull/201 do what you need?

yangbongsoo:
yes. I close this PR.

PR #206, opened 2020-07-03
Do not lcase element or attribute names that match SVG or MathML name…

### LINKED ISSUE #182: Preserve the casing of tags and attributes names.
It would be nice if the the policy builder had an option that disabled the "sanitization" of tags and attributes names.
Currently all names are converted to lowercase which is ok when you're using it for HTML only, but if there is an SVG image nested inside the HTML it breaks.
For example, when `viewBox` attribute on is converted to `viewbox` and the image is not displayed correctly.

mikesamuel:
…s exactly

> Currently all names are converted to lowercase which is ok when
> you're using it for HTML only, but if there is an SVG image nested
> inside the HTML it breaks.  For example, when `viewBox` attribute is
> converted to `viewbox` the image is not displayed correctly.

This commit splits *HtmlLexer*.*canonicalName* into variants which preserve
items on whitelists derived from the SVG and MathML specifications, and
adjusts callers of *canonicalName* to use the appropriate variant.

Fixes #182

@wookie41 @zeeneir do you have an example that I can turn into a unit test?

### DIFF
--- a/src/main/java/org/owasp/html/ElementAndAttributePolicies.java
+++ b/src/main/java/org/owasp/html/ElementAndAttributePolicies.java
@@ -35,7 +35,7 @@ import javax.annotation.concurrent.Immutable;
 
 /**
  * Encapsulates all the information needed by the
- * {@link ElementAndAttributePolicySanitizerPolicy} to sanitize one kind
+ * {@link ElementAndAttributePolicyBasedSanitizerPolicy} to sanitize one kind
  * of element.
  */
 @Immutable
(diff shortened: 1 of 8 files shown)


zeeneir:
Simple test for viewBox:

```java
    PolicyFactory policyFactory = new HtmlPolicyBuilder()
            .allowElements("svg")
            .allowAttributes("viewBox").onElements("svg")
            .toFactory();
    String svg = "<svg viewBox=\"0 0 0 0\"></svg>";
    assertEquals(svg, policyFactory.sanitize(svg));
```

mikesamuel:
@zeeneir Added tests.  Let me know if this PR solves your problem.

zeeneir:
Looks good to me. Thanks a lot!

Issue #213, opened 2020-11-26
possible xss attack in StandardUrlAttributePolicy


saaspeter:
if user input sting is: 

```html
<body>aaa bbb<a href=\"jav&#97script:alert(1)\">test1</a></body>
```

policy defined as this:

```java
PolicyFactory LINKS_RAW = (new HtmlPolicyBuilder()).allowElements("a").allowStandardUrlProtocols().allowAttributes("href","target").onElements("a").toFactory()
```

when check the "a" tag attribute href,  StandardUrlAttributePolicy will be used to the the href value, this method:  apply(String elementName, String attributeName, String value) will be called, but since there is '#' in the value, so it skipped the protocol check, and finally return the input href value, so even the value which not contain any protocol in "HTTP, https, mailto" , will not be filtered.

![image](https://user-images.githubusercontent.com/3772321/100342024-56e30900-3018-11eb-96c4-e6f9c8edf8bc.png)

is this a bug?

yangbongsoo:
@saaspeter Hello. I am sanitizer user.
I tested what you mentioned.

```java
  @Test
  public void testIssue213() {
    PolicyFactory policyFactory = new HtmlPolicyBuilder()
            .allowElements("a")
            .allowStandardUrlProtocols()
            .allowAttributes("href", "target")
            .onElements("a")
            .toFactory();
    String sanitize = policyFactory.sanitize("<a href=\"jav&#97script:alert(1)\">test1");
    System.out.println(sanitize); // <a href="jav&amp;#97script:alert%281%29">test1</a>
```
the result string is `<a href="jav&amp;#97script:alert%281%29">test1</a>`
even if href attribute value is remained, that string can not possible attack.
sanitizer transform `&` to `&amp;`


![스크린샷 2020-11-30 오후 4 17 16](https://user-images.githubusercontent.com/7060510/100579647-e42fa180-3327-11eb-8f6c-133487f7362c.png)

mikesamuel:
Thanks for the report.  Please do try to report bypasses as described in the issue template

>.   Please report **security vulnerabilities** via [OWASP's vulnerability rewards program](https://bugcrowd.com/owaspjavasanitizer).

I believe @yangbongsoo is correct.  `&#97` is missing a `;` so is not an HTML character reference, but if it were, it would be decoded to `'a'` before `apply` is called.

saaspeter:
[...]

PR #221, opened 2021-01-10
check html special letter in StandardUrlAttributePolicy

yangbongsoo:
@saaspeter @mikesamuel 

how about check the html special letters ex `&#97`


mikesamuel:
I thought issue #213 was not an actual vulnerability.

yangbongsoo:
I agree with you.
I wanted to show you that this PR code can filter it.
But I don't think this task is very necessary either.

I'll close this PR.

Issue #223, opened 2021-01-19
Forbidden numeric character references appear in sanitized HTML


simon-greatrix:
The HTML living standard ( https://html.spec.whatwg.org/multipage/syntax.html#character-references ) states:

> The numeric character reference forms described above are allowed to reference any code point excluding U+000D CR, noncharacters, and controls other than ASCII whitespace.

However, non-characters from the supplemental planes are encoded numerically.

The code:

```
    StringBuilder builder = new StringBuilder();
    Encoding.encodeRcdataOnto(Character.toString(0x5fffe), builder);
    System.out.println(builder.toString());
```

Produces: "&amp;#x5fffe;"

I see two possible simple possible solutions, but I am loathe to recommend either one:

First the characters could be elided in line with the elision of U+FFFE and U+FFFF. This produces a strange botch that is not required by the rules of HTML nor XML, and I don't like it.

Alternatively, the character is allowed if it is not numerically escaped and the noncharacters U+FDD0 to U+FDEF are presented unescaped - so consistency with other non-characters would produce legal HTML. However, all other supplemental code points are represented by numeric escapes to avoid corruption when converting between unicode encodings. I am not happy with introducing special cases for these supplemental code points.

More complex solutions would be to introduce a policy for handling of the "discouraged" characters defined in https://www.w3.org/TR/2008/REC-xml-20081126/#charsets. 

I am happy to put the time into creating a fix and test cases, but I need guidance as to what is the "correct" solution.

mikesamuel:
Thanks for the detailed report.

What about replacing any banned code-point with the replacement character (U+FFFD)?

It looks like, from that link, that the set of banned codepoints is
[U+0..U+8, U+B, U+C, U+E, U+F, U+7F..U+9F, U+FDD0..U+FDEF, U+FFFE, U+FFFF, U+1FFFE, U+1FFFF, U+2FFFE, U+2FFFF, U+3FFFE, U+3FFFF, U+4FFFE, U+4FFFF, U+5FFFE, U+5FFFF, U+6FFFE, U+6FFFF, U+7FFFE, U+7FFFF, U+8FFFE, U+8FFFF, U+9FFFE, U+9FFFF, U+AFFFE, U+AFFFF, U+BFFFE, U+BFFFF, U+CFFFE, U+CFFFF, U+DFFFE, U+DFFFF, U+EFFFE, U+EFFFF, U+FFFFE, U+FFFFF, U+10FFFE, U+10FFFF]
[...]

PR #225, opened 2021-02-07
Proposed fix for issue #223: forbidden character references in sanitized html

simon-greatrix:


src/main/java/org/owasp/html/Encoding.java
```diff
@@ -108,13 +111,16 @@ private static void stripBannedCodeunits(StringBuilder sb, int start) {
           if (i+1 < n) {
             char next = sb.charAt(i+1);
             if (Character.isSurrogatePair(ch, next)) {
-              sb.setCharAt(k++, ch);
-              sb.setCharAt(k++, next);
+              // The last two code points in each plane are non-characters that should be elided.
```
simon-greatrix, line 114: Added check for the supplementary plane non-characters

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -108,13 +111,16 @@ private static void stripBannedCodeunits(StringBuilder sb, int start) {
...
+              if ((ch & 0xfc3f) != 0xd83f || (next & 0xfffe) != 0xdffe) {
+                sb.setCharAt(k++, ch);
+                sb.setCharAt(k++, next);
+              }
               ++i;
             }
           }
           continue;
-        } else if ((ch & 0xfffe) == 0xfffe) {
+        } else if ((ch & 0xfffe) == 0xfffe || (0xfdd0 <= ch && ch <= 0xfdef)) {
```
simon-greatrix, line 123: Added check for BMP non-characters

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -240,127 +262,94 @@ private static void encodeHtmlOnto(
       char ch = plainText.charAt(i);
       if (ch < REPLACEMENTS.length) {  // Handles all ASCII.
         String repl = REPLACEMENTS[ch];
-        if (ch == '{' && repl == null) {
-          if (i + 1 == n || plainText.charAt(i + 1) == '{') {
-            repl = braceReplacement;
+        if( repl==null ) {
+          if (ch == '{') {
```
simon-greatrix, line 266: A none-replacement could still be a new-line needing normalization.

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -240,127 +262,94 @@ private static void encodeHtmlOnto(
...
+              // CRLF, so remove CR
+              repl = "";
+            }
           }
         }
         if (repl != null) {
           output.append(plainText, pos, i).append(repl);
           pos = i + 1;
         }
-      } else if ((0x93A <= ch && ch <= 0xC4C)
```
simon-greatrix, line 252: I removed the 2018 i-OS crashing test because:
(a) it should be patched in devices by now
(b) it was always legal HTML
(c) there was no test for the 2020 i-OS crashing flag text
(d) I do not believe it is this libraries job to catch all device specific risks

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -240,127 +262,94 @@ private static void encodeHtmlOnto(
...
-          char next;
-          if (i + 1 < n
-              && Character.isSurrogatePair(
-                  ch, next = plainText.charAt(i + 1))) {
-            // Emit supplemental codepoints as entity so that they cannot
-            // be mis-encoded as UTF-8 of surrogates instead of UTF-8 proper
-            // and get involved in UTF-16/UCS-2 confusion.
-            int codepoint = Character.toCodePoint(ch, next);
-            output.append(plainText, pos, i);
+      } else if (RISKY_NORMALIZATION.contains(ch)) {
```
simon-greatrix, line 287: New check which expands on the idea that U+1FEF normalized to back-tick. We now catch all unicode characters than have a compatibility decomposition that is even slightly risky.

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -240,127 +262,94 @@ private static void encodeHtmlOnto(
...
+          if( (codepoint & 0xfffe) != 0xfffe ) {
             appendNumericEntity(codepoint, output);
-            ++i;
-            pos = i + 1;
-          } else {
-            output.append(plainText, pos, i);
-            // Elide the orphaned surrogate.
-            pos = i + 1;
           }
-        } else if (0xfe60 <= ch) {
```
simon-greatrix, line 307: This case is now handled by the risky normalization set check.

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -240,127 +262,94 @@ private static void encodeHtmlOnto(
...
+
+  /**
+   * Append a codepoint to the output as a numeric entity.
+   *
+   * @param codepoint the codepoint
+   * @param output    the output
+   *
+   * @throws IOException              if the output cannot be written to
+   * @throws IllegalArgumentException if the codepoint cannot be represented as a numeric escape.
+   */
```
simon-greatrix, line 330: Note that this now throws an IllegalArgumentException if the numeric code point is forbidden in HTML.

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -240,127 +262,94 @@ private static void encodeHtmlOnto(
...
-                     ? codepoint < 0x100 ? 2 : 3
-                     : (codepoint < 0x10000 ? 4
-                        : codepoint < 0x100000 ? 5 : 6));
+      // Append a hexadecimal value
       output.append('x');
-      for (int digit = nDigits; --digit >= 0;) {
-        int hexDigit = (codepoint >>> (digit << 2)) & 0xf;
-        output.append(HEX_NUMERAL[hexDigit]);
-      }
+      output.append(Integer.toHexString(codepoint));
```
simon-greatrix, line 348: This code is simpler. Java is guaranteed to use lower case ASCII hex characters to represent the code point, so there seems no need for bespoke code.

src/main/java/org/owasp/html/Encoding.java
```diff
@@ -391,11 +380,34 @@ static void appendNumericEntity(int codepoint, Appendable output)
...
   static {
     for (int i = 0; i < IS_BANNED_ASCII.length; ++i) {
       IS_BANNED_ASCII[i] = !(i == '\t' || i == '\n' || i == '\r');
     }
   }
 
+  /** Set of all Unicode characters which when processed with unicode compatibility decomposition will include a non-alphanumeric ascii character. */
+  static final Set<Character> RISKY_NORMALIZATION;
+  static {
+    HashSet<Character> set = new HashSet<Character>();
```
simon-greatrix, line 393: For how these code-points were identified, see the unit-tests where the full scan of the BMP is explained and verified against the set here. The code was written this way for fast start up instead of doing the scan of the entire BMP.

src/main/java/org/owasp/html/HtmlLexer.java
```diff
@@ -527,7 +527,7 @@ private HtmlToken parseToken() {
             break;
           }
         }
-      } else if (!Character.isWhitespace(ch)) {
+      } else if (!isAsciiWhitespace(ch)) {
```
simon-greatrix, line 530: HTML requires ASCII whitespace, not Unicode whitespace.

src/test/java/org/owasp/html/SanitizersTest.java
```diff
@@ -313,7 +313,10 @@ public static final void testScriptInTable() {
       .and(Sanitizers.STYLES)
       .and(Sanitizers.IMAGES)
       .and(Sanitizers.TABLES);
-    assertEquals("<table></table>Hallo\r\n\nEnde\n\r", pf.sanitize(input));
```
simon-greatrix, line 316: new line normalization changed the expected output

src/test/java/org/owasp/html/HtmlSanitizerTest.java
```diff
@@ -392,53 +392,6 @@ public static final void testNbsps() {
             codeUnits));
   }
 
-  @Test
```
simon-greatrix, line 395: No longer supported. See my previous comment for Encoding.java

src/test/java/org/owasp/html/EncodingTest.java
```diff
@@ -305,4 +371,66 @@ void testBadlyDonePostProcessingWillnotAllowInsertingNonceAttributes()
     Encoding.encodeHtmlAttribOnto("a nonce=xyz ", attrib);
     assertEquals("a nonce&#61;xyz ", attrib.toString());
   }
+
+  @Test
+  public static final void testRiskyNormalizationSetContents() {
```
simon-greatrix, line 376: This unit test does the complete scan of the BMP for characters with a compatibility decomposition that could have side-effects.

src/test/java/org/owasp/html/ElidedCharactersTest.java
```diff
@@ -0,0 +1,143 @@
...
+ * unit tests verifies that the inclusion of such characters does not allow dangerous code to slip through.
+ * <p>
+ * There are two requirements:
+ * <p>
+ * 1) The Encoding.encodeRcdataOnto method should remove discouraged characters.
+ * 2) Sanitized HTML should not change
+ *
+ * @author Simon Greatrix on 25/01/2021.
+ */
+public class ElidedCharactersTest extends TestCase {
```
mikesamuel, line 22: I'll look at this PR in more detail shortly.  In the meantime, these tests seem to be failing on Travis.  Does that differ from what you see running locally?

src/main/java/org/owasp/html/Encoding.java
mikesamuel: I'm ok with this because of (a), but disagree on (d) because preventing denial of service is in scope for this project.  For the record, (c) was probably an oversight on my part.

mikesamuel:
I'll look at this in more detail soon.  Thanks so much.

simon-greatrix:
When you asked I was like "I would NEVER submit a merge request with failing test!". Then I looked at my IDE and saw that it said "No tests found", because it has forgotten how JUnit 4 works apparently. So, I'll have to change my claim to "I would never KNOWINGLY submit a merge request with failing tests!"

src/main/java/org/owasp/html/Encoding.java
simon-greatrix: If these magic character sequences are in scope, I think it would be excellent if they could be handled via a configuration file so that they can be quickly fixed. In production systems I've worked on, changing configuration is generally something that can be done faster than changing code, because the scope of potential regression issues is much smaller with a configuration file.

mikesamuel:
Heh.  Yeah, the codebase has some Java5/6 compatibility baggage.

simon-greatrix:
Looks OK now - hope those were the correct changes.

jmanico:
Reviving this. Issue #223 is still real and this is still the only fix proposed for it, and @mikesamuel's 2021 review was positive. The branch conflicts with `main` for structural reasons rather than logic: the sources moved under `owasp-java-html-sanitizer/`, `parent/pom.xml` and `scripts/build_for_travis.sh` no longer exist, and `ElidedCharactersTest` / `EncodingTest` have drifted. @simon-greatrix, if you are still willing, a rebase onto current `main` (dropping the parent pom and Travis script changes) would let the new PR build run and I will review and merge promptly. If I do not hear back in a few weeks I will try to carry the rebase myself, preserving your authorship.

jmanico:
Carried onto current main as #413, with your authorship preserved on the first commit and the review adjustments described there. Thank you for the fix and the tests, @simon-greatrix. I will close this PR once #413 merges.

jmanico:
Merged via #413, which carries these changes onto current main with your authorship on the first commit. Thanks again, @simon-greatrix.

Issue #237, opened 2021-10-27
Attributes allowed globally together with "style" are lost


corebonts:
Since the commit 020d5d0d7b8e985be32d3608612a9889135ef060 all attributes that are allowed globally are ignored, if "style" is given as the first attribute.

Problematic code:
```
    public HtmlPolicyBuilder globally() {
      if(attributeNames.get(0).equals("style")) {
        return allowStyling();
      } else {
        return HtmlPolicyBuilder.this.allowAttributesGlobally(
            policy, attributeNames);
      }
    }
```

Proof
 ```
 @Test
  public static final void testStyleWithOtherAttributesGlobally() {
    PolicyFactory policyBuilder = new HtmlPolicyBuilder()
            .allowAttributes("style", "align").globally()
            .allowElements("a", "label", "h1", "h2", "h3", "h4", "h5", "h6")
            .toFactory();
    String input = "<h1 style=\"color:green ;name:user ;\" align=\"center\">This is some green text</h1>";
    String want = "<h1 style=\"color:green\" align=\"center\">This is some green text</h1>";
    assertEquals(want, policyBuilder.sanitize(input));
  }
```
_Note that `align="center"` is missing from the output._

I will file a PR to fix the issue

sgesin:
Another problem that we're seeing with 020d5d0 is that it assumes the `attributeNames` list is non-empty. The proposed change in b493617 would help prevent having to write empty list checks prior to calling `globally`.

corebonts:
That's true. I got some comments on my pull request but I will try to adjust it today so hopefully these things will be fixed soon.

jmanico:
Fixed by #248 (the `contains("style")` change in `AttributeBuilder.globally()`), which shipped in 20240325.1. The test case from this issue is being added to `SanitizersTest` as a regression guard, credited to @corebonts. Closing.

Issue #234, opened 2021-09-01
CSS grids not supported


CGjupoulton:
Even with `allowStyling()`, this: 

```html
<div style="display: grid; grid-template-columns: repeat( auto-fit, minmax(160px, 1fr) );">
```

becomes 
```html
<div>
```
It seems that "display: grid" and all the other CSS grid properties are not part of the list of valid properties. How can I allow these?

mukuldhariwal94:
I think you need to do 
```
.allowAttributes(
                "style", "class"
        ).globally()
```

if you want to allow the style tag globally in all html elements or 

.allowAttributes(
                "style", "class"
        ).onElements("div");

CGjupoulton:
> if you want to allow the style tag in only the div elements

That works for other CSS properties such as color, but not the newer grid properties.

CGjupoulton:
`display: grid;` and `grid-template-columns: repeat( auto-fit, minmax(160px, 1fr) );` is the CSS I am trynig to allow. 

The `grid-template-columns` propery is quite complex, I'm unsure how to edit the CssSchema to allow it.

csware:
cf. https://github.com/OWASP/java-html-sanitizer/issues/192#issuecomment-1918722606 for a (manual) workaround

jmanico:
Thanks for the report, and apologies for the long wait.

Your exact value now works, but it needs an opt-in, and the reason is worth explaining because @mikesamuel's reply above was pointing at it.

`CssSchema.DEFAULT` withholds **the whole layout family** — `display`, `position`, `float`, `clear`, `overflow`, `z-index`, `opacity`, `visibility` and the offsets. It is not that `grid` was missing from `display`'s values; `display:block` is dropped exactly as `display:none` is. That is deliberate: a style attribute may restyle content, but it may not reposition it, cover the page around it, or hide it. Adding grid to DEFAULT would have quietly given every existing user of this library layout control over untrusted markup.

So #427 defines the grid family (and flexbox, `gap`, alignment, `transform`, and the `fr` unit) but leaves DEFAULT alone. A policy that wants layout opts in:

```java
CssSchema layout = CssSchema.withProperties(Arrays.asList(
    "display", "grid-template-columns", "grid-template-rows",
    "grid-column", "grid-row", "gap",
    "repeat()", "minmax()"));   // functions need listing too
PolicyFactory p = new HtmlPolicyBuilder()
    .allowElements("div").allowAttributes("style").globally()
[...]

PR #238, opened 2021-10-27
Do not ignore attributes allowed globally together with 'style' (#237)

corebonts:
Also, allowStyling() internally allows the 'style' attribute, so it is not necessary to ignore it.


mikesamuel:
Thanks for adding a testcase.  The check for style as the zero-th element seems good to change, but what prompted this?

Instead of using null for the policy, and checking `== null`, can we check for the identity policy?  My vague recollection was that `.join` was pretty good about just returning *x* when joining *x* with the identity policy.

corebonts:
First, thanks for reviewing it.
But sorry, it's not clear, what prompted what? The change that calls `allowStyling()` when "style" property is allowed or the change that uses now the contains check instead of the zero-th element check?

For the first, I don't know, it's someone else's change and I don't know the reason behind it. For me it also feels a bit magical.
For the latter one, we had failing tests in our product when we updated to the latest sanitizer.

And for your comment about the nullcheck, you're right, I will change that.

src/main/java/org/owasp/html/HtmlPolicyBuilder.java
```diff
@@ -888,7 +888,11 @@ public final class AttributeBuilder {
      * transformation by a previous policy.
      */
     public AttributeBuilder matching(AttributePolicy attrPolicy) {
-      this.policy = AttributePolicy.Util.join(this.policy, attrPolicy);
+      if (this.policy == null) {
+        this.policy = attrPolicy;
+      } else {
+        this.policy = AttributePolicy.Util.join(this.policy, attrPolicy);
+      }
```
mikesamuel, line 895: Does https://github.com/OWASP/java-html-sanitizer/pull/248/files#diff-a27b541fc6864e5b794ba42fc4230501e1fa203e2bd05cf782c52a44b1b4b54d in another PR fix this?

jmanico:
Thanks @corebonts, and sorry this sat for so long. The bug you reported in #237, other global attributes being dropped when `style` was first in the list, was fixed by #248, which replaced the early-return on `attributeNames.get(0)` with a `contains("style")` check that still falls through to `allowAttributesGlobally` for every name. Your `testStyleWithOtherAttributesGlobally` test case is being added to `SanitizersTest` on `main` under your authorship as a regression guard.

The second part of this PR, letting `matching(IDENTITY_ATTRIBUTE_POLICY)` bypass the CSS schema so `display: grid` passes raw, is the same request as #334 and we are declining it for the reasons written there: `style` always goes through the schema. Grid support belongs in `CssSchema` itself (#234). Closing as fixed.

Issue #247, opened 2022-01-24
Array out of bounds exception on HtmlPolicyBuilder initialization


mymhealthltd-joshengland:
Constructing a HtmlPolicyBuilder zero-defined global attributes, globally, leads to an Arry out of bounds exception

Here is sample code that produces the problem

```java
new HtmlPolicyBuilder().allowElements().allowAttributes().globally().toFactory();
```

The exception comes form an un-guarded check on the zeroth element of the attributesNames list. In this situation, attribute names is empty and so has no zeroth element.

```java
    public HtmlPolicyBuilder globally() {
      if(attributeNames.get(0).equals("style")) {
        return allowStyling();
      } else {
        return HtmlPolicyBuilder.this.allowAttributesGlobally(
            policy, attributeNames);
      }
    }
```

This construction used to work in version 20180219.1 of this library but is broken in 20211018.2.

jmanico:
Backlog triage ahead of the next release.

Verified fixed on `main`. `new HtmlPolicyBuilder().allowElements().allowAttributes().globally().toFactory()` no longer throws; there is a regression test for it in `HtmlPolicyBuilderTest`. Thanks for the clear report and the root-cause analysis.

PR #248, opened 2022-01-24
Fix allowAttributes().globally() (#247)

mymhealthltd-joshengland:
Add guard to .globally() method of HtmlPolicyBuilder to prevent
ArrayOutOfBoundsException when checking to see if the zeroth
element of the attributeNames list contains 'style'.

This restores behaviour present in version 202180219.1 which
allowed for an empty allowed attributes names list to be
specified globally through the builder.


mikesamuel (approved):

mikesamuel:
Thanks for the bug and the fix.

mikesamuel:
Fixes #300

PR #267, opened 2022-06-09
add inAttribute and change parameter order

yangbongsoo:
First of all, thank you for continuously doing the version up.

1. Added comment for newly added inAttribute in appendDecodedEntity method.
2. Like the order of the appendDecodedEntity method parameter, I modified it because I thought it would be more natural to have the html parameter come first. Since the inAttribute parameter is used first inside mayComplete method, it can be said that it was declared first, so if you say no, I will roll back.


jmanico:
Merged via #386, rebased onto the current module layout with your authorship preserved, @yangbongsoo. Thank you, and sorry it took four years.

PR #272, opened 2022-09-02
Remove Guava dependency and update to Java 9

### LINKED ISSUE #157: All-in-one version with shaded guava
Hi,

I'd like to use owasp-java-html-sanitizer in PrimeFaces (a popular JSF component suite). However, because of very restrictive policies regarding the use of third-party policies my PR cannot be merged.

The problem is the dependency to guava, which is a really big one that is also widely used and therefore version conflicts are conceivable.

So I come to the question if it would be possible for you to provide an additional all-in-one version of owasp-java-html-sanitizer having the guava dependency shaded?

Please see https://github.com/primefaces/primefaces/issues/3214 for the reasons why my PR was reverted.

Thanks.

### LINKED ISSUE #162: get rid of guava and use jdk8 as a minimum.
From my perspective there is zero need for a Guava dependency. Streams, try-with-resources, StandardCharsets would most likely already eliminate 90% of the usages.

Later version of JDK is also fine by me.

`<plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-compiler-plugin</artifactId>
          <version>3.3</version>
          <configuration>
            <source>6</source>
            <target>6</target>
          </configuration>
        </plugin>`

Targetting JDK6 implies you're ok with people still using JRE6 which is discontinued and full of security bugs.

claudioweiler:
This PR remove Guava dependency and update Java target/source to 9.

Please note that:

1. ~Two annotations from Guava are copied to source:~
   - ~`@GwtCompatible`: need only for `@VisibleForTesting`.~
   - ~`@VisibleForTesting`: used in 2 internal methods that should be visible only for test scope. Maybe there are a better approach.~
1. ~Replaced Guava `Preconditions` with `assert`. I'm sure there are a better approach, but `assert` is already used in some places, so I stick with that.~
1. ~Java sets do not ensures order, so unit testing eventually fail in `HtmlPolicyBuilderTest` when checking ordered `noopener noreferrer`, then just run again. Maybe review unit test.~
1. Added `configuration` to `maven-bundle-plugin` because `maven-verifier-plugin` fails.

Fixes #157, fixes #162
Also fixes rebuilding this library, caused by CVEs found on Guava, with just version update.

---

**Edit**

1. Rebased with last commits;
1. Guava annotations removed;
1. Some `Set`'s replaced with `List` to ensure proper ordering of `noopener noreferrer`.

---

**Edit 2**

Replaced Guava `Preconditions` with control structures and explicit throws.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -88,8 +88,11 @@ application while protecting against XSS.
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <guava.version>30.1-jre</guava.version>
+    <maven.compiler.source>9</maven.compiler.source>
+    <maven.compiler.target>9</maven.compiler.target>
+    <maven.compiler.release>9</maven.compiler.release>
   </properties>
+
 
   <build>
     <pluginManagement>
@@ -184,8 +187,8 @@ application while protecting against XSS.
           <artifactId>maven-compiler-plugin</artifactId>
           <version>3.3</version>
           <configuration>
-            <source>6</source>
-            <target>6</target>
+            <source>9</source>
+            <target>9</target>
           </configuration>
         </plugin>
         <plugin>
@@ -235,14 +238,9 @@ application while protecting against XSS.
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>com.google.guava</groupId>
-        <artifactId>guava</artifactId>
-        <version>${guava.version}</version>
-      </dependency>
-      <dependency>
         <groupId>commons-codec</groupId>
         <artifactId>commons-codec</artifactId>
-        <version>[1.4,)</version>
+        <version>[1.15,)</version>
       </dependency>
       <dependency>
         <groupId>com.google.code.findbugs</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -60,6 +60,11 @@
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <Export-Package>org.owasp.html</Export-Package>
+          </instructions>
+        </configuration>
       </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
@@ -81,10 +86,6 @@
   </build>
 
   <dependencies>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-    </dependency>
     <dependency>
       <groupId>commons-codec</groupId>
       <artifactId>commons-codec</artifactId>
--- a/src/main/java/org/owasp/html/AttributePolicy.java
+++ b/src/main/java/org/owasp/html/AttributePolicy.java
@@ -28,9 +28,8 @@
 
 package org.owasp.html;
 
-import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableList;
-
+import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.annotation.CheckReturnValue;
@@ -91,11 +90,11 @@ import org.owasp.html.Joinable.JoinHelper;
       }
 
       @Override
-      Optional<ImmutableList<AttributePolicy>> split(AttributePolicy x) {
+		Optional<List<AttributePolicy>> split(AttributePolicy x) {
         if (x instanceof JoinedAttributePolicy) {
           return Optional.of(((JoinedAttributePolicy) x).policies);
         } else {
-          return Optional.absent();
+			return Optional.empty();
         }
       }
 
--- a/src/main/java/org/owasp/html/CssSchema.java
+++ b/src/main/java/org/owasp/html/CssSchema.java
@@ -28,17 +28,17 @@
 
 package org.owasp.html;
 
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.TreeSet;
 
 import javax.annotation.Nullable;
-
-import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 
 /** Describes the kinds of tokens a CSS property's value can safely contain. */
 @TCB
@@ -54,11 +54,11 @@ public final class CssSchema {
     /** A bitfield of BIT_* constants describing groups of allowed tokens. */
     final int bits;
     /** Specific allowed values. */
-    final ImmutableSet<String> literals;
+	final Set<String> literals;
     /**
      * Maps lower-case function tokens to the schema key for their parameters.
      */
-    final ImmutableMap<String, String> fnKeys;
+	final Map<String, String> fnKeys;
 
     /**
      * @param bits A bitfield of BIT_* constants describing groups of allowed tokens.
@@ -66,11 +66,11 @@ public final class CssSchema {
      * @param fnKeys Maps lower-case function tokens to the schema key for their parameters.
      */
     public Property(
-        int bits, ImmutableSet<String> literals,
-        ImmutableMap<String, String> fnKeys) {
+        int bits, Set<String> literals,
+        Map<String, String> fnKeys) {
       this.bits = bits;
-      this.literals = literals;
-      this.fnKeys = fnKeys;
+      this.literals = Set.copyOf(literals);
+      this.fnKeys = Map.copyOf(fnKeys);
     }
 
     @Override
@@ -125,11 +125,11 @@ public final class CssSchema {
   static final int BIT_UNICODE_RANGE = 128;
 
   static final Property DISALLOWED = new Property(
-      0, ImmutableSet.<String>of(), ImmutableMap.<String, String>of());
+      0, Collections.emptySet(), Collections.emptyMap());
 
-  private final ImmutableMap<String, Property> properties;
+  private final Map<String, Property> properties;
 
-  private CssSchema(ImmutableMap<String, Property> properties) {
+  private CssSchema(Map<String, Property> properties) {
     if (properties == null) { throw new NullPointerException(); }
     this.properties = properties;
   }
@@ -144,14 +144,14 @@ public final class CssSchema {
    */
   public static CssSchema withProperties(
       Iterable<? extends String> propertyNames) {
-    ImmutableMap.Builder<String, Property> propertiesBuilder =
-        ImmutableMap.builder();
+    Map<String, Property> propertiesBuilder =
+        new HashMap<>();
     for (String propertyName : propertyNames) {
       Property prop = DEFINITIONS.get(propertyName);
       if (prop == null) { throw new IllegalArgumentException(propertyName); }
       propertiesBuilder.put(propertyName, prop);
     }
-    return new CssSchema(propertiesBuilder.build());
+    return new CssSchema(Map.copyOf(propertiesBuilder));
   }
 
   /**
@@ -161,8 +161,8 @@ public final class CssSchema {
    */
   public static CssSchema withProperties(
       Map<? extends String, ? extends Property> properties) {
-    ImmutableMap<String, Property> propertyMap =
-        ImmutableMap.copyOf(properties);
+    Map<String, Property> propertyMap =
+        new HashMap<>();
     // check that all fnKeys are defined in properties.
     for (Map.Entry<String, Property> e : propertyMap.entrySet()) {
       Property property = e.getValue();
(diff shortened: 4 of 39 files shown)


eivinhb:
Since Guava has a [security issue](https://github.com/google/guava/issues/4011) now, we are also looking into owasp html sanitizer getting rid of Guava. 
But releasing in java-html-sanitizer on java 9 is a no go for us. I guess that 8 should be used for some more years since java 8 is LTS and still get security updates. 
With regards to GwtCompatible and VisibleForTesting I believe that this compatibility should be removed all together. It is way to vendor specific imho.

eivinhb:
Yeah. Maby one of the big contributers (@mikesamuel for instance?) can say something about this. I am only a consumer but I can also contribute if we can agree on a direction. :)

eivinhb:
I do see the point in having proper immutable collections. Maybe use Apache collections4 for that? It has a much lower impact being a smaller more 'do one thing only'-library. That might be a solution until march 31 2025 when java 8 receives no further public updates.

claudioweiler:
For sure immutable collections have its usage. But immutable collections are needed only when you receive a collection from external usage, as you can't ensure proper collection handling.
IMHO, simple handling with unmodifiable and copies when necessary are enough to this library usage, but this need a deep look on all usages to ensure this behavior.
I take a quick look at collections4, and can't find immutable collections. Will try to create a simple test case.

talios:
_If_ the base version is going to move from Java 8, to Java 9 - can we please ensure that this is released as a major breaking version.

mikesamuel:
Thanks for diving into the codebase, @claudioweiler 

----

> Java sets do not ensures order

I thought LinkedHashSet did.

https://docs.oracle.com/javase/7/docs/api/java/util/LinkedHashSet.html
> This implementation differs from HashSet in that it maintains a doubly-linked list running through all of its entries. This linked list defines the iteration ordering, which is the order in which elements were inserted into the set (insertion-order).

Another difference between Guava immutable collections and Java standard library ones is that the Guava ones reject `null` at construction time, so consumers know they're not going to see `null` when iterating keys/values/elements.  Are there places where we might need additional `null` checks if we do this?

----

 > Two annotations from Guava are copied to source

I'd rather avoid this.  Copying files may break JAR sealing.
I suspect that, outside Google, no-one uses annotation processing to prevent references from non-test code that don't have private visibility so it only serves as documentation.
So one alternative to `@VisibleForTesting` is `// Only visible for testing`.

----

> If the base version is going to move from Java 8, to Java 9 - can we please ensure that this is released as a major breaking version.

iirc, the base version is Java 6.  I think there's some uses of reflection to paper over the nastiest Java6/Java8 problems.

https://github.com/OWASP/java-html-sanitizer/blob/33d319f876abbb35cb95eddf9705c46bd96822bd/src/main/java/org/owasp/html/AutoCloseableHtmlStreamRenderer.java#L15-L26

I started this project before semver was a thing.  I think there was some discussion of the difficulties that caused on #162.
Afaict, the only way to a modern versioning scheme would be to transition to a new maven artifact name which would prevent updates from automatically reaching existing users.

Maybe the way to resolve whether to target 8 or 9 or 11 and how to deal with versioning problems would be to post an announcement and send a message to the mailing list like "speak now if you need JRE<=8 support or forever hold your peace" and alert people that an artifact name &| versioning scheme change is coming down the pipe.

mikesamuel:
> That might be a solution until march 31 2025 when java 8 receives no further public updates.

My reading of the [JAVA SE LTS Roadmap](https://www.oracle.com/java/technologies/java-se-support-roadmap.html) is that JDK 8 is under support until 2030 at least.  Am I missing something?

> ### Oracle Java SE Support Roadmap
> | Release | GA Date | Premier Support Until | Extended Support Until | Sustaining Support |
> | -- | -- | -- | -- | -- |
> | 7 (LTS) | July 2011 | July 2019 | July 2022 | Indefinite |
> | 8﻿ (LTS) | March 2014 | March 2022 | **December 2030** | Indefinite |


>

mikesamuel:
> I take a quick look at collections4, and can't find immutable collections.

I don't think they use the term "immutable", and I'm not very familiar with Apache Collections, but perhaps @eivinhb was referring to [*Unmodifiable\**](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/Unmodifiable.html)

> Interface Unmodifiable
>
> All Known Implementing Classes:
[UnmodifiableBag](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/bag/UnmodifiableBag.html), [UnmodifiableBidiMap](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/bidimap/UnmodifiableBidiMap.html), [UnmodifiableBoundedCollection](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/collection/UnmodifiableBoundedCollection.html), [UnmodifiableCollection](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/collection/UnmodifiableCollection.html), [UnmodifiableEntrySet](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/map/UnmodifiableEntrySet.html), [UnmodifiableIterator](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/iterators/UnmodifiableIterator.html), [UnmodifiableList](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/list/UnmodifiableList.html), [UnmodifiableListIterator](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/iterators/UnmodifiableListIterator.html), [UnmodifiableMap](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/map/UnmodifiableMap.html), [UnmodifiableMapEntry](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/keyvalue/UnmodifiableMapEntry.html), [UnmodifiableMapIterator](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/iterators/UnmodifiableMapIterator.html), [UnmodifiableMultiSet](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/multiset/UnmodifiableMultiSet.html), [UnmodifiableMultiValuedMap](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/multimap/UnmodifiableMultiValuedMap.html), [UnmodifiableNavigableSet](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/set/UnmodifiableNavigableSet.html), [UnmodifiableOrderedBidiMap](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/bidimap/UnmodifiableOrderedBidiMap.html), [UnmodifiableOrderedMap](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/map/UnmodifiableOrderedMap.html), [UnmodifiableOrderedMapIterator](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/iterators/UnmodifiableOrderedMapIterator.html), [UnmodifiableQueue](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/queue/UnmodifiableQueue.html), [UnmodifiableSet](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/set/UnmodifiableSet.html), [UnmodifiableSortedBag](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/bag/UnmodifiableSortedBag.html), [UnmodifiableSortedBidiMap](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/bidimap/UnmodifiableSortedBidiMap.html), [UnmodifiableSortedMap](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/map/UnmodifiableSortedMap.html), [UnmodifiableSortedSet](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/set/UnmodifiableSortedSet.html), [UnmodifiableTrie](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/trie/UnmodifiableTrie.html)
>
> public interface [Unmodifiable](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/src-html/org/apache/commons/collections4/Unmodifiable.html#line.36)
>
> Marker interface for collections, maps and iterators that are unmodifiable.

claudioweiler:
Hi @mikesamuel 

> I thought LinkedHashSet did.

Yeah, it does. But the original DEFAULT_RELS_ON_TARGETTED_LINKS Set is copied to others Sets and then it looses order. It demands a proper analysis of usages to add LinkedHashSet on all necessary places.

> Another difference between Guava immutable collections and Java standard library ones is that the Guava ones reject null at construction time, so consumers know they're not going to see null when iterating keys/values/elements. Are there places where we might need additional null checks if we do this?

Yeah, Java 9 immutable disallows nulls. In Java 8 it will need proper handling on this.

> I'd rather avoid this. Copying files may break JAR sealing.
I suspect that, outside Google, no-one uses annotation processing to prevent references from non-test code that don't have private visibility so it only serves as documentation.
So one alternative to @VisibleForTesting is // Only visible for testing.

As I don't know if there is any annotation processing I preferred not to mess with this. If it's ok I can remove this, much better.

> I don't think they use the term "immutable", and I'm not very familiar with Apache Collections, but perhaps @eivinhb was referring to [Unmodifiable*](https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/Unmodifiable.html)

The problem is that immutable and unmodifiable are different concepts. Unmodifiable collections do not protect from external changes and already exists on Java 8 native libraries.

If goal will be Java 8, then unmodifiable must be used, but with additional protections when creating/copying collections.

eivinhb:
> is that JDK 8 is under support until 2030 at least

I believe that you need to pay for support until 2030. But googling for resources about this is kinda confusing at the moment.

> The problem is that immutable and unmodifiable are different concepts. Unmodifiable collections do not protect from external changes and already exists on Java 8 native libraries.

Yeah, I realise that now. Maybe java 11 is the way to go then, or even java 17 since many major libraries like Spring is doing the huge leap i these days. I guess OWASP need to make some decisions about this.
Is it possible with the licences on Guava to copy the code for immutable collections and change the package name but keep the code unchanged? That will not break JAR sealing, true?

claudioweiler:
> Maybe java 11 is the way to go then, or even java 17 since many major libraries like Spring is doing the huge leap i these days.

Java 17 is a huge leap indeed!

I think Java 11 is a good step. But Java 8 is possible too, just need some more work.

> Is it possible with the licences on Guava to copy the code for immutable collections and change the package name but keep the code unchanged?

Guava uses Apache License 2.0: https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)
I'm not confident here, but I think that it fits in "modify" case. But, I think that this is going to be much more work than just adapting to Java 8.

Flame-Phoenix:
Any update about this pull request.

alinbrici46:
Anyone have update on when this will be fixed?

claudioweiler:
Removing Guava is a best way. Guava is a huge umbrella lib, that leads to constantly CVEs being found.

Also, current Guava version used by owasp-java-html-sanitizer is guava:30.1-jre, that is based on Java 8. Why this lib still on Java 6?

melloware:
Agreed but a short term fix without a lot of regression testing right now would be to update to 32.0.0.jre but I like this PR as the long term fix.

melloware (approved):

mikesamuel:
This was a lot of work.  Thanks for putting that in.

Is there a stdlib way to do something like Preconditions?  I worry that without  -ea asserts don't run and it's often better to fail than to return without upholding a security property.

claudioweiler:
> Is there a stdlib way to do something like Preconditions? I worry that without -ea asserts don't run and it's often better to fail than to return without upholding a security property.

The standard is `assert`. But, IMHO, I would go with simples `if...throw IllegalArgumentException...` as pointed in guava docs.

I can change all `assert` points if you agree.

melloware:
@claudioweiler i think we should go with simple `if` its cleaner code than all the `asserts` just my two cents.

claudioweiler:
Changing my assertions to control structures.

```
Preconditions.checkArgument > throw IllegalArgumentException
Preconditions.checkState > throw IllegalStateException
Preconditions.checkNull > throw NullPointerException
Preconditions.checkElementIndex > throw IndexOutOfBoundsException
```

There are still asserts in code. Those are original asserts and I choose not to mess with it, but I can do it with some directions.

claudioweiler:
I don't know how it happened, but my commits are with a merge commit in the middle, can anybody help me remove this?

mikesamuel:
I think fixing up other asserts is probably out of scope for this PR.

melloware (approved):

mikesamuel:
Thanks @claudioweiler for the patch, and for keeping with it through my delays.

melloware:
@mikesamuel is there any chance you can do a release?  Our security team still keeps dinging this library as having the Guava vulnerability.

PR #282, opened 2023-05-31
Guava 32.0.0-jre fixes multiple CVE's

melloware:
Fixes at least these two CVE's

```xml
    <suppress>
        <notes>
            <![CDATA[file name: guava.jar PrimeFaces does not use the temp directory functionality that is vulnerable]]></notes>
        <packageUrl regex="true">^pkg:maven/com\.google\.guava/guava@.*$</packageUrl>
        <cve>CVE-2020-8908</cve>
    </suppress>
    <suppress>
       <notes><![CDATA[file name: guava.jar PrimeFaces does not use the temp directory functionality that is vulnerable]]></notes>
       <packageUrl regex="true">^pkg:maven/com\.google\.guava/guava@.*$</packageUrl>
       <cve>CVE-2023-2976</cve>
    </suppress>
```

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -88,7 +88,7 @@ application while protecting against XSS.
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <guava.version>30.1-jre</guava.version>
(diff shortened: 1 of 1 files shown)


tandraschko:
can't we just remove guava?

melloware:
Yep there is another PR here to remove it entirely. I just put this up in case they wanted a quick fix before a more thorough review of the other PR

Issue #251, opened 2022-02-07
Stripping off the contents when Child Combinator are present in the media queries 


log2akshat:
Hi,

We are using this library in [Zimbra](https://github.com/Zimbra/java-html-sanitizer-release-20190610.1) for sanitization of the e-mail body and during sanitization of the customer-generated HTML, we came across the following situation when we have `Child Combinator` in the media queries and during sanitization, it is not able to parse properly and the whole HTML is stripped off. 

```
@media yahoo {
               .d-mobile {
                  display: none !important;
               }

               .d-desktop {
                  display: block !important;
               }

               .w-lg-75>tbody>tr>td {
                  width: 75% !important;
               }

               .w-lg-100 {
                  width: 100% !important;
               }

               .w-lg-100>tbody>tr>td {
                  width: 100% !important;
               }

               .pt-lg-0>tbody>tr>td,
               .py-lg-0>tbody>tr>td {
                  padding-top: 0 !important;
               }

               .pr-lg-0>tbody>tr>td,
               .px-lg-0>tbody>tr>td {
                  padding-right: 0 !important;
               }

               .pb-lg-0>tbody>tr>td,
               .py-lg-0>tbody>tr>td {
                  padding-bottom: 0 !important;
               }

               .pl-lg-0>tbody>tr>td,
               .px-lg-0>tbody>tr>td {
                  padding-left: 0 !important;
               }
.....
```
It works fine if I pull the `Child Combiator` out of the media query I got the following message but it is able to perform sanitization without stripping off the contents.

```
Child:[style: null]
Next Child:null
```

It will be great if someone can guide me on how to handle this situation or it can be considered as an enhancement or bugfix.

log2akshat:
It will be great if I can get some regarding this issue.
Thanks

uttamtakalkar:
OWASP Team, any updates on this issue ?

uttamtakalkar:
@jmanico , can you please look into this issue

uttamtakalkar:
@jmanico any updates on this ?

jmanico:
I'm honestly not sure. cc @mikesamuel could you take a look at this when time allows?

log2akshat:
@uttamtakalkar I think we should stop highlighting and raising a particular issue for fixing, especially personally tagging the developers of that project.

[...]

PR #297, opened 2023-12-20
Fix : CSS Child Combinator Parsing Bug

subbudvk:
https://github.com/OWASP/java-html-sanitizer/commit/241b4b8a6cd37c486b36f12fc135a018881c3e4a#diff-0a08f29a5b7867e56d6aa9f6abe035e32ee9411a8bc96afa9a6acff2a6d6f07fR338 

The above commit was made to make parsing consistent with HTML5 Spec, but while this is being rewritten it looks like a regression was introduced when parsing for `>`

To ensure this entity `>` is part of a HTML Comment it seems to have been checked the previous two characters are `-` but with this commit only `char - 2` is checked twice. So if a CSS Child combinator with `-` selector is used, it was treated was a error and as a result user CSS was badly stripped. Fixing this and adding a test for it. 

Fixes https://github.com/OWASP/java-html-sanitizer/issues/251

### DIFF
--- a/src/main/java/org/owasp/html/HtmlStreamRenderer.java
+++ b/src/main/java/org/owasp/html/HtmlStreamRenderer.java
@@ -341,7 +341,7 @@ public class HtmlStreamRenderer implements HtmlStreamEventReceiver {
           }
           break;
         case '>':
-          if (i >= 2 && sb.charAt(i - 2) == '-' && sb.charAt(i - 2) == '-') {
+          if (i >= 2 && sb.charAt(i - 2) == '-' && sb.charAt(i - 1) == '-') {
             if (innerStart < 0) { return i - 2; }
             // Merged start and end like <!--->
             if (innerStart + 6 > i) { return innerStart; }
--- a/src/test/java/org/owasp/html/HtmlPolicyBuilderTest.java
+++ b/src/test/java/org/owasp/html/HtmlPolicyBuilderTest.java
@@ -994,6 +994,43 @@ public class HtmlPolicyBuilderTest extends TestCase {
     assertEquals("x<textArea>y</textArea>", textAreaPolicy.sanitize(input));
   }
 
+ @Test
+  public static final void testCSSChildCombinator() {
+	 HtmlPolicyBuilder builder = new HtmlPolicyBuilder();
+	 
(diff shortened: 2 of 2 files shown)


subbudvk:
@mikesamuel  @jmanico

src/main/java/org/owasp/html/HtmlStreamRenderer.java
```diff
@@ -341,7 +341,7 @@ private static int checkHtmlCdataCloseable(
           }
           break;
         case '>':
-          if (i >= 2 && sb.charAt(i - 2) == '-' && sb.charAt(i - 2) == '-') {
+          if (i >= 2 && sb.charAt(i - 2) == '-' && sb.charAt(i - 1) == '-') {
```
mikesamuel, line 344: Thanks

src/test/java/org/owasp/html/HtmlPolicyBuilderTest.java
```diff
@@ -994,7 +994,7 @@ public static final void testTextareaIsNotTextArea() {
     assertEquals("x<textArea>y</textArea>", textAreaPolicy.sanitize(input));
   }
 
-   @Test
+  @Test
```
mikesamuel, line 997: This is just an opportunistic fix while merging.

src/test/java/org/owasp/html/HtmlPolicyBuilderTest.java
```diff
@@ -1007,6 +1007,41 @@ public static final void testCSSFontSize() {
...
+ 	 	  + "width:100%;\n"
+ 	 	  + "}\n"
+ 	 	  + "\n"
+ 	 	  + ".hdg-1>._inner {\n"
+ 	 	  + "background-color: #666;\n"
+ 	 	  + "}\n"
+ 	 	  + "-->\n"
+ 	 	  + "</style>";
+ 	  assertEquals(toSanitize, factory.sanitize(toSanitize));
+  }
```
mikesamuel, line 1044: Thanks for the unit test.

mikesamuel (approved):

log2akshat:
@subbudvk Thanks for fixing the bug.

PR #299, opened 2024-01-15
Fix index out of bound exception when attributes list is empty

1003n40:
Fixes: #300


mikesamuel:
Thanks for the fix.  Would this other PR, https://github.com/OWASP/java-html-sanitizer/pull/248/files#diff-a27b541fc6864e5b794ba42fc4230501e1fa203e2bd05cf782c52a44b1b4b54d , fix that?

1003n40:
Hi @mikesamuel, as I checked further the other PR it would fix that. As of my point of view I didn't see it earlier so decided to make a fix for that, but if it is already created just close the issue I created with it, and we could close that PR.

On the other hand, I have another question, when new release come for that library, as of now I can only see the last one is from the year 2022.

mikesamuel:
I'm pushing a new release shortly.

Issue #301, opened 2024-01-23
Guava removal breaks compatibility (with JDK9)


csware:
The minimum required Java version was explicitly raised to 9 with the recent removal of Guava (cf. commit 3b6cc1b7e6d7992c5b15f87f91dc163de7d35c05).

I appreciate that Guava was dropped recently, however, there is a problem:
[`Map.copyOf`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#copyOf(java.util.Map)) (also `List.copyOf`) was first introduced with Java 10.

The method [`Files.readString`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#readString(java.nio.file.Path,java.nio.charset.Charset)) used in the tests, was introduced with Java 11 (here, [Commons-IO](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FileUtils.html#readFileToString(java.io.File,java.nio.charset.Charset)) could be used, or just plain `new String(Files.readAllBytes(path), Charsets.toCharset(charset));`).

I suppose reasons are:
1. The compiler plugin is too old and does not support the ``--release`` parameter (cf. #304)
2. No automatic checks are automatically executed (cf. #305)

csware:
PR #302 recodes the usage of `*.copyOf`, still there are some places where it is still used.

Should we introduce our own helper method?

csware:
Btw. for me the following tests fails:
```
org.owasp.html.HtmlPolicyBuilderTest
testRelLinksWhenRelisPartOfData(org.owasp.html.HtmlPolicyBuilderTest)
junit.framework.AssertionFailedError: Failure in testRelLinksWhenRelisPartOfData
```

melloware:
I am not sure Java9 should be supported as its a Non LTS version: https://www.oracle.com/java/technologies/java-se-support-roadmap.html

I think Java 11 LTS is the lowest version that should be supported

csware:
Many Apache Common projects still target 1.8 that's why I'm asking. Otherwise, we should target Java 11.

csware:
Here you can find a proof of concept for supporting Java 1.8: https://github.com/csware/java-html-sanitizer/tree/target-jdk8

melloware:
I like the idea of supporting JDK8

csware:
[...]

PR #302, opened 2024-01-23
Avoid unnecessary copies

csware:
These copies were introduced with PR #272.

This related to #301 and works with Java 9.


src/main/java/org/owasp/html/CssSchema.java
```diff
@@ -846,15 +846,15 @@ Property forKey(String propertyName) {
     builder.put("z-index", bottom);
     builder.put("repeating-linear-gradient()", linearGradient$Fun);
     builder.put("repeating-radial-gradient()", radialGradient$Fun);
-	DEFINITIONS = Map.copyOf(builder);
+    DEFINITIONS = Collections.unmodifiableMap(builder);
```
mikesamuel, line 849: Ok.  I see builder allocated locally at 263

mikesamuel (approved):

src/main/java/org/owasp/html/CssSchema.java
csware: Exactly, I renamed all variables to include bulder to make it more clear.

csware:
SHall I rebase?

PR #303, opened 2024-01-23
Remove mentionings of guava 

csware:


mikesamuel (approved):

csware:
@mikesamuel 
I'm not really happy, because you squared my commits and GitHub used "csware" instead of my real name. Can we add a .mailmap?

PR #311, opened 2024-01-28
Undo JDK6 hack

csware:
Undo AutoClosable hack introduced for JDK6 compatibility.

This is safe to do, as it was added in JDK7 and the Java-HTML-Sanitizer now requires JDK9 (even requiring 8 would still be ok)

This basically reverts commit f7e09249ad65b8f5a276e3fdd079f4c48b6eef79 (and commit 74c6dd0247ea8c676aa8ce9db8c1497db2763aa5).


melloware (approved):

src/test/java/org/owasp/html/AutoCloseableHtmlStreamRendererTest.java
```diff
@@ -1,85 +0,0 @@
...
-
-    assertFalse(closeable.closed);
-
-    AutoCloseableHtmlStreamRenderer.closeIfAnyCloseable(closeable);
-
-    assertTrue(closeable.closed);
-  }
-
-  @Test
-  public static void testThatAutoCloseablesAreClosed() throws IOException {
```
mikesamuel, line 44: Do we have any other tests that closing an autocloseable closes the underlying appendable?

csware: THe original code (to which I reverted), also had to tests.

csware: How is this expected to work?

Take the following code:
```Java
package org.owasp.html;

import java.io.IOException;

import junit.framework.TestCase;

@SuppressWarnings("javadoc")
public class AutoCloseableHtmlStreamRendererTest extends TestCase {
	  public static void testBlup() {
	    PolicyFactory policyBuilder = new HtmlPolicyBuilder()
	      .allowAttributes("href").onElements("a")
	      .allowStandardUrlProtocols()
	      .allowElements("a")
	      .toFactory();

	    AutoClosableAppender autoClosableAppender = new AutoClosableAppender();
	    assertFalse(autoClosableAppender.closed);
		HtmlStreamRenderer renderer = HtmlStreamRenderer.create(autoClosableAppender,
				Handler.DO_NOTHING,
				Handler.DO_NOTHING);
		// Use the policy defined above to sanitize the HTML.
		HtmlSanitizer.sanitize("<a href=http://example.com>A link</a>", policyBuilder.apply(renderer));
		assertTrue(autoClosableAppender.closed);
	  }

	private static class AutoClosableAppender implements Appendable, AutoCloseable {
		public boolean closed = false;
		@Override
		public void close() throws Exception {
			closed = true;
		}

		@Override
		public Appendable append(CharSequence csq) throws IOException {
			return this;
		}

		@Override
		public Appendable append(CharSequence csq, int start, int end) throws IOException {
			return this;
		}

		@Override
		public Appendable append(char c) throws IOException {
			return this;
		}
	}
}
```

Close is never called (tested with: e8aa0f1b75ab81eea97ef173fefc023a3e16963d (current main) and eda1118b8c48f9e04ee916bec49102851bcfe943)
Also HtmlStreamRenderer.CloseableHtmlStreamRenderer.close seems to be never called

mikesamuel: Yeah.  Java has intersection types but doesn't allow mentioning them.

mikesamuel (approved):

src/test/java/org/owasp/html/AutoCloseableHtmlStreamRendererTest.java
csware: If this is the intended usage, maybe we need another factory method and make the inner-classes public

PR #312, opened 2024-01-30
Add overflow-wrap to CssSchema definition list

### LINKED ISSUE #290: CSS property `overflow-wrap` not included in CssSchema definition list
The `CssSchema` class provides [a map of CSS property names and values](https://github.com/OWASP/java-html-sanitizer/blob/main/src/main/java/org/owasp/html/CssSchema.java#L254-L847). This list is extensive, but missing some of the more recent additions to CSS.

I would appreciate considering [the `overflow-wrap` property](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) for inclusion in this list. This CSS property directs browsers on how to insert line breaks when text would overflow its box. If the maintainers find this agreeable, I can provide a PR for consideration.

csware:
Fixes #290


src/main/java/org/owasp/html/CssSchema.java
```diff
@@ -423,6 +423,8 @@ Property forKey(String propertyName) {
         "auto", "inherit", "none");
     Set<String> overflowLiterals0 = Set.of(
         "auto", "hidden", "inherit", "scroll", "visible");
+    Set<String> overflowWrapLiterals0 = Set.of(
+        "normal", "break-word", "anywhere", "inherit");
```
csware, line 427: What about other valid values such as `"initial" , "revert" , "revert-layer", "unset"` ((listed as "global values" on https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap?

mikesamuel: I think revert and revert-layer were added before the last time this property list was thoroughly revamped.  I don't quite understand the semantics, but I would worry about adding it everywhere that `inherit` appears in case it allows working around visual containment and event capturing measures like clipping, position, and overflow.

Issue #192, opened 2020-02-03
word-break in style is discarded because it is not considered a valid value in CSSSchema


mrabhishek:
input:

```html
<div>
<table>
<td style="word-break: keep-all;">
</td>
</table>
</div>
```

policy:
`Sanitizers.BLOCKS
                .and(Sanitizers.FORMATTING)
                .and(Sanitizers.LINKS)
                .and(Sanitizers.TABLES)
                .and(Sanitizers.IMAGES)
                .and(Sanitizers.STYLES)
                .and(.and(new HtmlPolicyBuilder()
                        .allowElements("style")
                        .allowAttributes("style").onElements("td", "table","div")
                        .allowAttributes("type", "word-break").onElements("style")
                        .toFactory());`

Expected output (should contain word-break).
```html
<div>
<table>
<td style="word-break: keep-all;">
</td>
</table>
</div>
```

juanmacoo:
`word-wrap` does not have any similar behaviour to `break-all`. Is there a reason it is not included in the allowed attributes?

mrabhishek:
How do we get an answer for this one? It does not look like the default Style policy can be overridden - If not, then it means that there is no way to provide a custom CSS schema that can allow elements like `word-break` and `display` that are not part of the default CSS schema. 

Looking for some explanation on why "display" is part of CSS definitions but not in the default schema that is used in Style,

https://github.com/OWASP/java-html-sanitizer/blob/main/src/main/java/org/owasp/html/CssSchema.java#L593

csware:
The default CSS definitions can be overwritten (inprinciple, but currently it does not work because of issue #313), by adding the following code to the HTMLPolicyBuilder:
```Java
              .allowStyling(
                  CssSchema.withProperties(
                      Map.of("word-break",
                          new CssSchema.Property(0,
                              Set.of("keep-all", "valid-values"),
                              Collections.emptyMap()))))
```

jmanico:
Backlog triage ahead of the next release.

Verified against `main`:
```java
new HtmlPolicyBuilder().allowElements("td")
    .allowAttributes("style").onElements("td")
    .allowStyling(CssSchema.withProperties(List.of("word-break")))
    .toFactory()
    .sanitize("<td style=\"word-break: keep-all;\">x</td>");
```
[...]

PR #315, opened 2024-01-31
Update valid word-break CSS property values

csware:
Based on https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

What about the generic values such as `inherit`, `initial`, etc.`?

refers to issue #192


mikesamuel:
> Based on https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
> 
> What about the generic values such as `inherit`, `initial`, etc.`?

Seems fine.

> refers to issue #192

mikesamuel:
What is the relationship between `word-wrap` and `word-break`?

csware:
> What is the relationship between `word-wrap` and `word-break`?

Ah, great catch! I changed the wrong literals...

Updated.

mikesamuel (approved):

PR #318, opened 2024-02-02
Get tests passing and recognize foreign content nodes better

mikesamuel:
Prepping for a release.

One test was not passing:
HtmlBuilderTest.testRelLinksWhenRelisPartOfData

I was also getting IDE warnings that there were some uses of `@since 10` APIs which made it hard to run tests outside mvn.

This commit cleans up those issues and a few warnings about `@deprecated` in javadoc without a corresponding annotation.

### DIFF
--- a/src/main/java/org/owasp/html/Encoding.java
+++ b/src/main/java/org/owasp/html/Encoding.java
@@ -43,6 +43,7 @@ public final class Encoding {
    * @return text/plain
    * @deprecated specify whether s is in an attribute value
    */
+  @Deprecated
   public static String decodeHtml(String s) {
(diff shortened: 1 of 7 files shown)


csware:
Some of these fixes are also on my PRs... Would be nice of you could comment on issue #301

csware:
FOr a new release you should also consider PR #313

subbudvk:
@mikesamuel : May I know when we are planning to have a new release, changes from some of our PRs are needed for us

subbudvk:
@mikesamuel can we get a release please with new fixes?

PR #319, opened 2024-02-02
Drop travis badge and add GitHub workflow badge

csware:
What about the Coverage badge? It is also outdated...


mikesamuel:
Ugh.  Everything related to coverage tracking and reporting ends up harder to integrate than it seems it should be.

mikesamuel (approved):

PR #321, opened 2024-02-02
Fix JDK9

csware:
This make the build work with JDK9

Another PR for JDK8 is on the way soon if you are interested...

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -179,17 +179,13 @@ application while protecting against XSS.
                 <value>bar</value>
               </property>
             </javaApiLinks>
-            <source>6</source>
+            <source>${maven.compiler.source}</source>
           </configuration>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>3.3</version>
-          <configuration>
-            <source>9</source>
-            <target>9</target>
-          </configuration>
+          <version>3.12.1</version>
         </plugin>
         <plugin>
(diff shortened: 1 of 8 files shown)


mikesamuel:
There was a lot of discussion on PR #272 about which version to support and the difference between immutable (unqualified ["unmodifiable" in Java SE parlance](https://docs.oracle.com/javase/10/docs/api/java/util/Collection.html#unmodifiable)) and unmodifiable view collections when moving away from Guava's immutable collections.

iirc, the decision to go with JDK 10 (we thought they were 9 which is on me) was that the *.of* and *.copyOf* factories recognize each others outputs and so one can pass a collection through multiple layers that defensively copy without incurring the cost of those copies multiple times.

I have no opinion on JDK 9 vs 10.  They both are post-modules.  I'd be happy to just go straight from JDK6 support in the last release to JDK 10.

csware:
For me moving to 10 would be fine, however, apache common mostly still support JDK8. The question is what JDKs are targeted by the users of java-html-sanitizer (cf. https://github.com/OWASP/java-html-sanitizer/issues/301#issuecomment-1920343365)

csware:
All my (private) projects are using JDK17, but as a library I suppose you have to swallow a pill and consider JDK8 (until there is enough pressure for everyone to use modern JDKs)...

csware:
It would also be possible to implement our own read-only wrapper for Map, Set and List and easily get "one copy", but is it worth this for people who use such old JDKs? Also, the builder shouldn't be used that frequently.

csware:
Any news on this?

PR #326, opened 2024-02-27
Support standard attributes in TABLES Policy

### LINKED ISSUE #195: Allow colspan attribute on td and th elements in TABLES PolicyFactory
It looks like the `TABLES` PolicyFactory in [Sanitizers](/OWASP/java-html-sanitizer/blob/master/src/main/java/org/owasp/html/Sanitizers.java) class does not allow the colspan attribute on `td` and `th` elements.

I see no reason why it couldn't.

subbudvk:
Fixes https://github.com/OWASP/java-html-sanitizer/issues/195 and allows more attributes than existing PR

The goal is to allow common safe attributes in table context for in a single PR instead of handling one attribute at a time. Pl let me know if we want to allow more, or if there are issues with this one.

### DIFF
--- a/src/main/java/org/owasp/html/Sanitizers.java
+++ b/src/main/java/org/owasp/html/Sanitizers.java
@@ -27,6 +27,9 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 package org.owasp.html;
+import java.util.Arrays;
+import java.util.List;
+
 
 /**
  * Pre-packaged HTML sanitizer policies.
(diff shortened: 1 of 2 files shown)


subbudvk:
@mikesamuel

jmanico:
I am sorry, @mikesamuel is the release master. He'll get on this as soon as he can. Forgive the delay!

ssweber:
Hello. What is the current thinking on this? Is "colspan","rowspan" generally safe to allow?

jmanico:
PR #378 has just been merged and adds `colspan` and `rowspan` (integer-checked) to `Sanitizers.TABLES`, which I believe covers the main ask here and the question from @ssweber. This branch now conflicts with `main`. If there are other table attributes you still want in the default policy, please rebase and trim this down to just those, with a test for each, and it can go in quickly. Otherwise I will close this as superseded in a couple of weeks. Thanks for the patience, @subbudvk.

jmanico:
Closing as superseded. The colspan/rowspan half landed via #378, and this branch still patches the pre-#379 `src/` layout, so it can no longer be rebased cleanly. The remaining two attributes, `headers` on td/th and `scope` on th, are landing in #407 on a fresh branch with positive and negative tests and a changelog credit to you. One change from this diff: `scope` is now emitted in canonical lower case rather than returning the original value after a lowercased comparison. Thanks for the original work, @subbudvk.

PR #327, opened 2024-03-01
Fixes in HTML Lexer to support HTML empty comment statements

mfrohike:
## Summary and intro

The HTML Lexer fails to detect empty HTML comment declarations leading into the next piece of HTML object in the input to not be detected as such. 

Basically, for an input such as `<!>`, the lexer considers the whole blob as an HTML comment instead of an empty comment declaration (`<!>`) and then an `img` tag (which is what browsers would do).

**This bug in the lexer, provides wrong information to the `HtmlChangeListener`**

## HTML RFC context

In the HTML RFC, comments are defined as:

```
To include comments in an HTML document, use a comment declaration. A comment declaration consists of `<!' followed by zero or more comments followed by `>'. Each comment starts with `--' and includes all text up to and including the next occurrence of `--'.
```

This means that `<!>` is a valid comment declaration with zero comments inside. Which means the following HTML code would trigger the `alert(1)` if rendered in a browser:

```html
<html><body>
<!><script>alert(1)</script>
</body></html>
```

In addition, the pattern `<!-->` is also transformed by browsers to `<!-- -->` (not sure why yet, but tested in all major browsers). This means the following HTML code would also trigger the `alert(1)` if rendered in a browser:

```html
<html><body>
<!--><script>alert(1)</script>
</body></html>
```

## The bug

The Lexer does not consider neither `<!>` nor `<!-->` as valid comment declaration statements, considering the last character of both statements (`>`) still as part of the comment. 
This means, when the sanitizer reads the following input `<!>`, the lexer will interpret **the whole input** as an HTML comment. 
However, the expected behavior would be to detect `<!>` as an HTML comment declaration, and `` as an `img` HTML tag.

This is easy to see with the `HtmlChangeListener` class. For example, the following code provides the following output:

```java
package org.owasp.html;

import java.util.ArrayList;
import java.util.List;

class BugTest
{
	public static void main(String[] args)
	{
		String[] test = new String[]{"qwe1<img>qwe2", "qwe3<!><img src=1>qwe4", "qwe5<!--><img src=1>qwe6"};
		List<String> results = new ArrayList<String>();
		for (String s : test)
		{
			MyListener htmlChangeListener = new MyListener();
			PolicyFactory sanitizePolicy = new HtmlPolicyBuilder().toFactory();
			String safeString = sanitizePolicy.sanitize(s, htmlChangeListener, results);
			System.out.println( "safeString :"  + safeString + " "+htmlChangeListener.getOutOfPolicyObjects());
		}

	}


}

... 

package org.owasp.html;

import javax.annotation.Nonnull;
import javax.annotation.Nullable;
import java.util.LinkedList;
import java.util.List;

public class MyListener implements HtmlChangeListener<List<String>>
{
	private final List<String> outOfPolicyObjects = new LinkedList<String>();

	@Override
	public void discardedTag(@Nullable List<String> context, @Nonnull String elementName)
	{
		outOfPolicyObjects.add(elementName);
	}

	@Override
	public void discardedAttributes(@Nullable List<String> context, @Nonnull String tagName, @Nonnull String... attributeNames)
	{
		outOfPolicyObjects.add(String.format("%s (%s)", tagName, String.join(",", attributeNames)));
	}

	public List<String> getOutOfPolicyObjects()
	{
		return outOfPolicyObjects;
	}
}
```

The output:

```
safeString :qwe1qwe2 [img]
safeString :qwe3qwe4 []
safeString :qwe5 []
```

However, the expected correct output should be:

```
safeString :qwe1qwe2 [img]
safeString :qwe3qwe4 [img]
safeString :qwe5qwe6 [img]
```

## The fix

We added a new state called `COMMENT_DASH_AFTER_BANG` into the `HtmlInputSplitter` inside `HtmlLexer` to handle dash character after the bang one (`!-`)
Also we created special condition checks in the `BANG` and `BANG_DASH` states inside the lexer's state machine  to handle `<!>` and  `<!-->` comments.

## Authors

**Bug discovery**: Carlos Villa (@carlosvillasanchez)
**Code fix**: Eduardo Aguado (@jfbyers)


subbudvk:
@jfbyers Can you provide a test case, where with a minimal policy, post sanitization, the vector is not sanitized?

jfbyers:
Hi @subbudvk , with the current implementation the following strings are not properly sanitized:

```java
class BugTest
{
	public static void main(String[] args)
	{
		String[] test = new String[]{"qwe1<img>qwe2", "qwe3<!>qwe4", "qwe5<!-->qwe6"};
		for (String s : test)
		{
			PolicyFactory sanitizePolicy = new HtmlPolicyBuilder().toFactory();
			String safeString = sanitizePolicy.sanitize(s, null, null);
			System.out.println( "safeString :"  + safeString);
		}

	}
}
``` 
This outputs:

```
safeString :qwe1qwe2
safeString :qwe3
safeString :qwe5
```

And it should be:

```
safeString :qwe1qwe2
safeString :qwe3qwe4
safeString :qwe5qwe6
```


This is even more misleading if you use a listener to add tags/attributes in
```java
discardedTag() 
````
or 

```java
discardedAttribute()
``` 

from the string being sanitized  e.g string `qwe3<!>  `   would **not** add `` as a discarded tag so any library consumer might take incorrect decisions about the contents of the string.

melloware (approved):

melloware (changes requested):
Agreed please add some more test cases

melloware (approved):

melloware:
No you are good.  I don't have commit privs so I can't run your workflow only @mikesamuel has permissions to this repo.

mikesamuel:
Which HTML RFC are you quoting?

https://html.spec.whatwg.org/multipage/syntax.html#comments seems to disagree. https://html.spec.whatwg.org/multipage/parsing.html#parse-errors **does** class `<!` under incorrectly opened comment, but why would the incorrectly opened comment end at `>` though instead of `-->`?

> And it should be:
>
> safeString :qwe1qwe2
> safeString :qwe3qwe4
> safeString :qwe5qwe6

Why?

src/main/java/org/owasp/html/HtmlLexer.java
```diff
@@ -640,17 +641,28 @@ && canonicalElementName(start + 2, end)
...
+                    state = State.COMMENT_DASH_AFTER_BANG;
                   } else {
                     state = State.DIRECTIVE;
                   }
                   break;
+                case COMMENT_DASH_AFTER_BANG:
+                  if ('>' == ch) { // <!--> is a valid html comment
+                    state = State.DONE;
+                    type = HtmlTokenType.COMMENT;
+                  }else{
```
mikesamuel, line 662: Spacing

src/test/java/org/owasp/html/HtmlLexerTest.java
```diff
@@ -121,6 +121,31 @@ public static final void testShortTags() {
...
+
+  @Test
+  public static final void testCommentDeclarationWith0CommentsAndTag() throws Exception
+  {
+    assertTokens("<!--><img>",
+            "COMMENT: <!-->",
+            "TAGBEGIN: <img",
+            "TAGEND: >"
+    );
+  }
```
mikesamuel, line 147: Thanks for the tests.

Could you add tests for 1 and 3 dashes?
They needn't be 3 tests.
You could probably incorporate them into this test by appending

    a<!--->b<!->c

subbudvk:
@mikesamuel : Can you kindly release a new version at least with whatever changes that are already merged to master?

melloware:
+1 to @subbudvk just get any new version out there that gets rid of Guava.

jfbyers:
Added support  for 2 comment parser errors  https://html.spec.whatwg.org/multipage/parsing.html#parse-errors :

- abrupt-closing-of-empty-comment  (i.e., ). The parser behaves as if the comment is closed correctly.
- incorrectly-closed-comment comment that is closed by the "--!>"  code point sequence. The parser treats such comments as if they are correctly closed by the "-->" code point sequence.

carlosvillasanchez:
> Which HTML RFC are you quoting?
> ....

Hey @mikesamuel

We are quoting this RFC: https://www.ietf.org/rfc/rfc1866.txt section 3.2.5.
 
>  To include comments in an HTML document, use a comment declaration. A comment declaration consists of `<!` followed by zero or more  comments followed by `>`. Each comment starts with `--` and includes  all text up to and including the next occurrence of `--`. In a  comment declaration, white space is allowed after each comment, but not before the first comment.  The entire comment declaration is ignored

Based on this, `<!>` it is indeed a valid self closed comment declaration. Without any comment.

It is true, that based on this, `<!-->` it is not a valid self closed comment declaration. However, we have seen in the most used browsers (Chrome, Firefox, Edge, Safari), the behavior is to modify this blob into `<!---->` which is indeed a valid self closed comment section, with one empty comment.

I can build some PoC for both scenarios detailed above, let me know if that is needed or if this is enough information.

Thanks!!

mikesamuel:
> +1 to @subbudvk just get any new version out there that gets rid of Guava.

Done

mikesamuel:
> We are quoting this RFC: https://www.ietf.org/rfc/rfc1866.txt section 3.2.5.

I don't think any modern browser references that standard and, since it's an RFC, it hasn't changes since it was published in 1995.
[The WhatWG steering group](https://github.com/whatwg/sg?tab=readme-ov-file#steering-group-representatives) is representatives from the main browser vendors, and html.whatwg.spec.org is the best approximation of what HTML is.

src/main/java/org/owasp/html/HtmlLexer.java
```diff
@@ -665,6 +679,8 @@ && canonicalElementName(start + 2, end)
                   if ('>' == ch) {
                     state = State.DONE;
                     type = HtmlTokenType.COMMENT;
+                  } else if ('!' == ch) {  // --!> is also valid closing sequence
+                    state = State.COMMENT_DASH_DASH;
```
mikesamuel, line 683: This would seem to suggest that`<!-- --!->` is a whole comment tag.
iiuc, after `<!-- --!-` we should be in the comment_dash state.

Perhaps we need `!` to transition to `COMMENT_DASH_DASH_BANG` here which transitions as does `case COMMENT` above.

mikesamuel: Does this look right?

```mermaid
flowchart TD
  BANG -- "-" --> BANG_DASH;
  BANG_DASH -- "-" --> COMMENT_DASH_AFTER_BANG;
  BANG_DASH -- "else" --> DIRECTIVE;
  COMMENT_DASH_AFTER_BANG -- "-" --> COMMENT_DASH_AFTER_BANG;
  COMMENT_DASH_AFTER_BANG -- ">" --> DONE;
  COMMENT_DASH_AFTER_BANG -- "else" --> COMMENT;
  COMMENT -- "-" --> COMMENT_DASH;
  COMMENT -- "else" --> COMMENT;
  COMMENT_DASH -- "-" --> COMMENT_DASH_DASH;
  COMMENT_DASH -- "else" --> COMMENT;
  COMMENT_DASH_DASH -- ">" --> DONE;
  COMMENT_DASH_DASH -- "-" --> COMMENT_DASH_DASH;
  COMMENT_DASH_DASH -- "!" --> COMMENT_DASH_DASH_BANG;
  COMMENT_DASH_DASH_BANG -- ">" --> DONE;
  COMMENT_DASH_DASH_BANG -- "-" --> COMMENT_DASH;
  COMMENT_DASH_DASH_BANG -- "else" --> COMMENT;
```

subbudvk:
> > +1 to @subbudvk just get any new version out there that gets rid of Guava.
> 
> Done

Thanks @mikesamuel ! Appreciate your contribution to the open source.

carlosvillasanchez:
> I don't think any modern browser references that standard and, since it's an RFC, it hasn't changes since it was published in 1995.
> [The WhatWG steering group](https://github.com/whatwg/sg?tab=readme-ov-file#steering-group-representatives) is representatives from the main browser vendors, and html.whatwg.spec.org is the best approximation of what HTML is.

About this, please forgive me in advance... I am not super used to referencing in RFCs and/or browsers standards.

That being said, in https://html.spec.whatwg.org/multipage/syntax.html#comments they state a comment must start with `<!--` and finish with `-->`, which as we are seeing it is not true. Browsers clearly identify a pattern like `<!>` as a comment section.

Not sure if this means this HTML spec is wrong, or it is just not the one followed by browsers. 

Also the pattern `<!-->` that browsers turn into `<!-- -->` its not specified in the spec (I guess this one rightfully so).

Again, let me know if it is useful to provide a live PoC for this.

src/main/java/org/owasp/html/HtmlLexer.java
carlosvillasanchez: Yes, this seems correct. The missing part is the `COMMENT_DASH_AFTER_BANG` case and the `<!>` scenario too. But the new `COMMENT_DASH_DASH_BANG` suggestion seems correct to me.

jfbyers: Thanks @mikesamuel , this makes sense. I implemented your proposed changes and added new tests. Everything passes the test. Can you please validate / approve the MR? Thanks!

carlosvillasanchez: @jfbyers Changes look good to me, let see what mike's thoughts are :)

jfbyers: @subbudvk   @mikesamuel  are you ok moving forward with this PR ? What are the next steps? Thank you.

jmanico:
Decision: merging this. The final version follows the WHATWG tokenizer, which is the reference this sanitizer tries to match, not RFC 1866: `<!>`, `<!-->` and `<!--->` are complete (empty) comments per the abrupt-closing-of-empty-comment and incorrectly-opened-comment parse errors, and `--!>` closes a comment per incorrectly-closed-comment. Browsers all agree on this today. Before this change the lexer swallowed everything after `<!-->` up to the next real `-->`, which is over-stripping rather than a bypass, but it still means the sanitizer disagreed with the browser about where a comment ends, and that is exactly the kind of gap we want closed.

I pushed one tiny commit adding the `@Test` annotation that `testDashDashBangComment` was missing, so it actually runs. Thanks for the persistence @mfrohike and @carlosvillasanchez, and @jtmelton for the nudge. It goes in when the PR build is green.

PR #218, opened 2020-12-12
Fixed allowAtributes("style").globally()

aakritisi:
Fixed #179

PR #334, opened 2024-04-11
allowAttributes("style").globally() shouldn't imply allowStyling() - Regression with 2024 version

### LINKED ISSUE #331: Issue in 2024x version with styles
https://github.com/OWASP/java-html-sanitizer/pull/218/files made a breaking change to imply allowAttributes("style").globally(), with this change I believe the CSS Schema based whitelisting is applied automatically even if someone explicitly use _allowAttributes()_ instead of _allowStyling()_

Though this change is made long back this may not had much impact for users who didn't done whitelisting previously because the change assumed the style would be in the first place of attribute list.

Now with https://github.com/OWASP/java-html-sanitizer/pull/248/files#diff-a27b541fc6864e5b794ba42fc4230501e1fa203e2bd05cf782c52a44b1b4b54d, this change is being forced so whitelisting gets automatically applied. 

We understand the use of whitelisting but this cannot be forced immediately to our users as we don't how much of user style is already allowed in CSS Schema. 

Can this be an option or be like existing where user have option to use allowStyling() or allowAttributes("style") without forcing this? Let me if you want this to be expressed API wise in different manner but still supporting two behaviours as it used to be, I'll submit a PR for this.

@mikesamuel

subbudvk:
This recent breaking changes

- Forces validating global style content with CSSSchema - Earlier we had better options seperately
    allowAttributes("style").globally() - doesn't sanitize, allowStyling() - did sanitize.
- make disallowAttribute("style").globally() now does allowStyling() as pointed out in the PR by someone. 

@mikesamuel @jmanico : Can you kindly have a look as we are facing issues after 2024 version upgrade.

Fixes https://github.com/OWASP/java-html-sanitizer/issues/331

### DIFF
--- a/owasp-java-html-sanitizer/src/main/java/org/owasp/html/HtmlPolicyBuilder.java
+++ b/owasp-java-html-sanitizer/src/main/java/org/owasp/html/HtmlPolicyBuilder.java
@@ -967,9 +967,6 @@ public class HtmlPolicyBuilder {
      */
     @SuppressWarnings("synthetic-access")
     public HtmlPolicyBuilder globally() {
-      if (attributeNames.contains("style")) {
-        allowStyling();
-      }
       return HtmlPolicyBuilder.this.allowAttributesGlobally(
           policy, attributeNames);
     }
--- a/owasp-java-html-sanitizer/src/test/java/org/owasp/html/SanitizersTest.java
+++ b/owasp-java-html-sanitizer/src/test/java/org/owasp/html/SanitizersTest.java
@@ -541,17 +541,6 @@ public class SanitizersTest extends TestCase {
             pf.sanitize(input)
     );
   }
-
-  @Test
-  public static final void testStyleGlobally() {
-    PolicyFactory policyBuilder = new HtmlPolicyBuilder()
-        .allowAttributes("style").globally()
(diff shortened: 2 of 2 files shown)


subbudvk:
@mikesamuel

csware:
Please also add test cases for the cases that must not happen.

subbudvk:
@mikesamuel

jmanico:
Decision: we are not going to take this, and I want to be straightforward about why.

Before #218, `allowAttributes("style").globally()` without `allowStyling()` passed the raw `style` value through untouched. That is an XSS-adjacent footgun in a sanitizer (CSS can carry `expression()`, `behavior:`, `-moz-binding`, `url()` beacons, and containment escapes), and #218 closed it by making any allowance of `style` run through the CSS schema. Reverting that would make the library unsafe by default for anyone who reaches for the obvious API. This PR also deletes the test that guards the behaviour and adds no negative tests, which @csware pointed out.

For the underlying need, "our users rely on CSS the default schema does not allow", the supported path is to widen the schema rather than bypass it:

```java
.allowStyling(CssSchema.union(CssSchema.DEFAULT,
    CssSchema.withProperties(Arrays.asList("grid-template-columns", ...))))
```

and per-element styling is being designed in #381. If a property you need cannot be expressed in `CssSchema`, please open an issue with the property and we will look at adding it. Closing this and #331.

PR #335, opened 2024-04-12
Allows start, end, justify-all and match-parent as text-align values.

martin-jackson:
After this change, all values currently listed on MDN for text-align are allowed: https://developer.mozilla.org/en-US/docs/Web/CSS/text-align#values

`left` and `right` are not in `textAlignLiterals0`, but they are already included through `azimuthLiterals1`:

https://github.com/OWASP/java-html-sanitizer/blob/f729a089b20aef49ed9ffd7ed1c7e207eee71dc5/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java#L315

https://github.com/OWASP/java-html-sanitizer/blob/f729a089b20aef49ed9ffd7ed1c7e207eee71dc5/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java#L714-L716

### DIFF
--- a/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java
+++ b/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java
@@ -452,7 +452,7 @@ public final class CssSchema {
     Set<String> tableLayoutLiterals0 = j8().setOf(
         "auto", "fixed", "inherit");
     Set<String> textAlignLiterals0 = j8().setOf(
-        "center", "inherit", "justify");
+        "center", "end", "inherit", "justify", "justify-all", "match-parent", "start");
     Set<String> textDecorationLiterals0 = j8().setOf(
         "blink", "line-through", "overline", "underline");
     Set<String> textTransformLiterals0 = j8().setOf(
--- a/owasp-java-html-sanitizer/src/test/java/org/owasp/html/HtmlPolicyBuilderTest.java
+++ b/owasp-java-html-sanitizer/src/test/java/org/owasp/html/HtmlPolicyBuilderTest.java
@@ -1295,6 +1295,23 @@ public class HtmlPolicyBuilderTest extends TestCase {
   }
 
   @Test
+  public final void testCSSTextAlign() {
+    HtmlPolicyBuilder builder = new HtmlPolicyBuilder();
+    PolicyFactory factory = builder.allowElements("span")
+        .allowAttributes("style").onElements("span").allowStyling()
+        .toFactory();
+
+    String toSanitizeTextAlignStart = "<span style=\"text-align:start\">start</span>";
+    assertEquals(toSanitizeTextAlignStart, factory.sanitize(toSanitizeTextAlignStart));
+
+    String toSanitizeTextAlignEnd = "<span style=\"text-align:end\">end</span>";
+    assertEquals(toSanitizeTextAlignEnd, factory.sanitize(toSanitizeTextAlignEnd));
+
+    String toSanitizeTextAlignFoo = "<span style=\"text-align:foo\">foo</span>";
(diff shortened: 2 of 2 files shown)


eivinhb (approved):
LGTM

martin-jackson:
@mikesamuel Hi! Does this look ok to merge and release? We need to allow the missing values, and I haven't found a way to override the existing `text-align` rule. The `union()` method of `CssSchema` throws an exception when trying to combine the default schema with my own schema:

```java
var updatedTextAlign = Map.of("text-align", new CssSchema.Property(0, Set.of("center", "end", "inherit", "justify", "justify-all", "match-parent", "start", "left", "right"), Collections.emptyMap()));
var schema = CssSchema.union(
    CssSchema.DEFAULT,
    CssSchema.withProperties(updatedTextAlign)
);
```

https://github.com/OWASP/java-html-sanitizer/blob/f729a089b20aef49ed9ffd7ed1c7e207eee71dc5/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java#L204-L207

EugenMayer:
2 years later, the spec that was violated here still is violated AND there are 2 new ones (left/right)

Could we please get this PR moving and merged - it is hard to argue that values that are part of the HTML spec are not part of the literals - this is a trivial PR.

Is this repository and project to be considered abandoned?

jmanico:
Merged. There are new maintainers. Are you interested in helping manage this repo? The original builders had to move on.

jmanico:
Here is the thread discussing thew new leadership team incase you care to help. https://github.com/OWASP/java-html-sanitizer/issues/360

EugenMayer:
@jmanico thank you for the help here. I'am not sure i can involve myself in good fait in an additional project (if the question was to my direction) in an more official capicity - i'am currently already involved in a couple of projects myself.

I can though dedicate time from time to time and help with what i have.

EugenMayer:
@jmanico the values you just merged are better, but nowdays outdated, see https://github.com/KontextWork/java-html-sanitizer/pull/1/changes#diff-576e2d03bc457727cc0e451d099b067f6e9a7ccf2dd5255cd9c3c7c7762e9d84R454

Should i create a new PR to add those missing or do you want to just add them yourself due to the triviality?

jmanico:
> @jmanico the values you just merged are better, but nowdays outdated, see https://github.com/KontextWork/java-html-sanitizer/pull/1/changes#diff-576e2d03bc457727cc0e451d099b067f6e9a7ccf2dd5255cd9c3c7c7762e9d84R454
> 
> Should i create a new PR to add those missing or do you want to just add them yourself due to the triviality?

How is this? https://github.com/OWASP/java-html-sanitizer/pull/420

EugenMayer:
perfect!

jmanico:
Thanks for the pointer to your fork — no need for a new PR, both are merged.

**#420** takes `textAlignLiterals0` to exactly the set in your diff: `center, end, inherit, justify, justify-all, match-parent, start, left, right, initial, revert, revert-layer, unset`. (`left`/`right` already reached `text-align` through the union with `azimuthLiterals1`, but they are spelled out now so the set reads for itself.)

**#421** then fixes the general case, which your diff made obvious: `text-align` was only one instance. `CssSchema` had 31 literal sets naming `inherit` and none naming `initial`/`revert`/`revert-layer`/`unset`, so `color: revert` and `font-size: unset` were dropped too. The CSS-wide keywords are now folded into *every* property rather than listed per-property. Function-argument schemas are deliberately excluded — `initial` is a value for `color`, not for a channel of `rgb(...)` — and inside a function it is dropped exactly as `foo` and `expression` already were.

The keywords are inert (no URLs, no functions, no new token classes), so this widens what round-trips, not what can execute.

Both are on `main` for the next release. And understood re: capacity — occasional help like this is genuinely useful, so thanks for pushing on it.

Issue #132, opened 2017-12-16
Sanitizing Html readers


mehretaban:
Sanitizer currently supports html string as input which can lead to a significant memory footprint for large html contents also large html contents could have been sent intentionally as a DOS attack for sanitizer. Supporting a Reader (or InputStream) input and buffering the portion of data  could be a useful feature.
Also (large) inner texts of text containers can be passed as separate readers which helps to implement truncation logic without significant effect on heap.

davidbarbrowatwork:
I have a need for this as well, and forked and created a PR for exactly this in the AntiSamy project. Would prefer using this library to AntiSamy. Happy to fork and put up a PR here if that would be welcome.

mikesamuel:
This would only need to affect the HTML lexer, but would be a pretty major change to it.  PRs welcome.

PR #339, opened 2024-04-19
Make StylingPolicy public

EugenMayer:
See #132 for the motivation.

Did not use a builder since having access to StylingPolicy.apply can be very critical


EugenMayer:
Any chance to have a opinion / review for this one? Thanks for the effort

EugenMayer:
Maybe we could at least get an answer on this this year? :) Thanks!

jmanico:
@EugenMayer sorry this sat unanswered for so long. The use case is a good one and we do want to take it — I've written up a proposed shape in #381 (a public factory on `CssSchema` rather than making `StylingPolicy` itself public, plus a note on how `url()` values must be handled outside the builder). If you're still interested, a revised PR against #381 with tests would be very welcome; otherwise we'll pick it up. Thanks for your patience.

PR #344, opened 2024-08-21
Rename ForJava9AndLater to ForJava10AndLater to match the actual meaning

csware:


jmanico:
Rebased this onto `main` for you, @csware, keeping the rename and the `catch (Throwable)` change from #343 together. Will merge once the build passes.

PR #351, opened 2024-10-14
Issue-338 Changing Java8Shim fallback to catch Throwable

katta524:
Java8Shim fallback expecting an Error, few implementations are throwing ClassNotFoundException for missing class ForJava9AndLater.


java8-shim/src/main/java/org/owasp/shim/Java8Shim.java
```diff
@@ -19,7 +19,7 @@ public abstract class Java8Shim {
             try {
                 // This is compiled with -release 1.9 in a separate project.
                 _instance = Class.forName("org.owasp.shim.ForJava9AndLater").newInstance();
-            } catch (Error e) {
+            } catch (Throwable e) {
```
katta524, line 22: Changing to Throwable, to ClassNotFoundException

Issue #361, opened 2025-08-22
How can we add CSS function calc() to CssChema


flashouti:
Hi! I want the CSS function [calc() ](https://developer.mozilla.org/es/docs/Web/CSS/calc) not to be deleted during sanitizing from "max-width" attribute. How can I add this finction to CssChema?

jmanico:
done

jmanico:
Landed in #362 (merge commit cacca13), so this is now genuinely done rather than just marked done. With the default schema, `calc()` is accepted in `width`, `min-width`, `max-width`, `height`, `min-height` and `max-height`. For example `max-width: calc(100% - 20px)` now survives sanitization as `max-width:calc( 100% - 20px )`.

Operands are limited to numbers, dimensions and percentages joined by `+`, `-`, `*`, `/` and parentheses. Anything else inside the call, including `var()`, `attr()` and `url()`, is stripped. If you build a custom `CssSchema` from property names, list `"calc()"` next to the sizing property, the same way `color` needs `rgb()`. This ships in the next release.

PR #362, opened 2025-11-23
Allow calc() in CSS schema and add to default whitelist

Raibipasha-24:
Allow calc() in CSS schema and add to default whitelist.

Adds a calc() function key to the CSS schema and includes it in the default whitelist. Follow-up needed to wire calc() into numeric property fnKeys and to add tests.
Files changed
[CssSchema.java] — add calc() to DEFINITIONS and DEFAULT_WHITELIST.

Can you review my pr and tell me?


jmanico (changes requested):
Thanks for taking this on, calc() is a real gap (#361). Two issues stop this from working as written, though:

1. **The new `max-width` entry is silently overwritten.** `builder` is a plain `HashMap`, and about 150 lines further down main already has `builder.put("max-width", maxHeight)`. That later put replaces the entry you add here, so `calc()` never becomes reachable from `max-width`. The fix is to change the existing `max-width` line rather than add a second one. You probably also want the same function map on `width`, `min-width`, `height`, `max-height`, and `min-height`, since those are where people actually use calc().

2. **Arithmetic operators are dropped.** `calc()` is defined as `new Property(1, j8().setOf(), zeroFns)`, so it accepts quantities but has an empty literal set. `StylingPolicy.punctuation()` only emits a token if it is in the literal set, so `calc(100% - 20px)` comes out as `calc(100% 20px)`, which is invalid CSS and the browser discards the whole declaration. The literal set needs `"+"`, `"-"`, `"*"`, `"/"` (and probably `"("`, `")"` for nested grouping). Note the security angle here: multiplication and division are safe in calc(), but please keep the operand types limited to quantities and percentages. No `var()`, no `attr()`, no `url()`.

3. **No test.** Please add a positive case (e.g. `width:calc(100% - 20px)` round-trips intact) and a negative case (e.g. `width:calc(expression(alert(1)))` or a `url()` inside calc is stripped) to `CssSchemaTest` or `HtmlSanitizerTest`. The project asks for both, see the Contributing section of the README.

Happy to merge once those are in.

jmanico:
Thanks for kicking this off. I pushed two commits onto your branch to finish it: a merge of `main`, and a follow-up that wires `calc(` into `width`, `min-width`, `max-width`, `height`, `min-height` and `max-height` (the `max-width` entry in the original commit was being overwritten by the existing `put` further down), gives `calc()` the `+ - * / ( )` literals so operators survive, keeps operands limited to quantities and percentages, and adds positive and negative tests plus a changelog entry. Your commit is preserved as the first commit of the PR. Merging now; this closes #361.

Issue #358, opened 2025-05-26
Form with <optgroup> incorrectly rewritten


baptistegrimaud:
When using [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/optgroup) within a `<select>` element, the sanitized HTML is invalid and contains extra `<select>` tags.
The following HTML:
```html
<form><select><optgroup label="mygroup"><option>My option</option></optgroup></select></form>
```
would become:
```html
<form><select><optgroup label="mygroup"><select><option>My option</option></select></optgroup></select></form>
```

Note that without the `<optgroup>` tag, the sanitized HTML is correct.
 
Unit test to reproduce the issue (with version [20240325.1](https://github.com/OWASP/java-html-sanitizer/releases/tag/release-20240325.1)):
```java
import org.junit.Test;
import org.owasp.html.HtmlPolicyBuilder;
import org.owasp.html.PolicyFactory;

import static org.junit.Assert.assertEquals;

public class OptgroupTest {

    @Test
    public void not_working() {
        HtmlPolicyBuilder policyBuilder = new HtmlPolicyBuilder();
        PolicyFactory factory = policyBuilder.allowElements("form", "select", "optgroup", "option").allowAttributes("label").globally().toFactory();
        String html = "<form><select><optgroup label=\"mygroup\"><option>My option</option></optgroup></select></form>";
        String result = factory.sanitize(html);
        assertEquals(html, result); // this fails!
//        Expected :<form><select><optgroup label="mygroup"><option>My option</option></optgroup></select></form>
//        Actual   :<form><select><optgroup label="mygroup"><select><option>My option</option></select></optgroup></select></form>
    }

    @Test
    public void working() {
        HtmlPolicyBuilder policyBuilder = new HtmlPolicyBuilder();
        PolicyFactory factory = policyBuilder.allowElements("form", "select", "option").toFactory();
        String html = "<form><select><option>My option</option></select></form>";
        String result = factory.sanitize(html);
        assertEquals(html, result);
    }
}
```

Is there anything missing when creating the `PolicyFactory` to properly support `<optgroup>`?

NB: Same problem seems to happen with `<datalist>` tags.

Shangeeth29:
Hi @baptistegrimaud 
I’ve opened a PR with the fix here: https://github.com/OWASP/java-html-sanitizer/pull/375

melloware:
@jmanico  this issue can be closed now that its merged

[...]

PR #375, opened 2026-01-06
Fix invalid nested <select> when sanitizing <optgroup>

Shangeeth29:
**Problem**
When using `<optgroup>` elements within `<select>` tags, the HTML sanitizer was incorrectly inserting extra `<select>` tags, producing invalid HTML.

**Input:**
<select><optgroup label="mygroup"><option>My option</option></optgroup></select>

**Broken Output**
<select><optgroup label="mygroup"><select><option>My option</option></select></optgroup></select>

**Root Cause**
Simple typo in HtmlElementTables.java

**Fixes** #358 

**Note**
Similar behavior was observed with <datalist> but is not included in this PR.


Shangeeth29:
Hi @jmanico 
Just a gentle nudge on this PR when you get a chance

jmanico:
cc @aalmiray @melloware (new leads)

melloware (approved):
Absolutely approve this is clearly wrong in the code and the correct fix

Issue #376, opened 2026-01-22
CVE-2025-66021 Mitigation Blocked Due to WebLogic 12 Class Loader Compatibility Issue


jguyea:
We've recently identified a compatibility issue while attempting to upgrade our java-html-sanitizer library from version 20220608.1 to 20260102.1. While upgrading to the latest version is necessary to address CVE-2025-66021, the new release introduces a problem with the WebLogic 12 class loader in our current Java 8 environment.

The issue stems from a Java catch clause in class Java8Shim introduced in release 20240325.1, which is incompatible with the WebLogic 12 class loader. As a result, the upgrade fails, blocking our ability to mitigate the CVE.

Proposed Solution:
It appears that this issue has already been addressed in the following pull request, though the fix has not yet been included in a formal release: #343

The single-line change in this commit resolves the incompatibility with the WebLogic 12 class loader. Specifically, the WebLogic class loader returns an Exception (ClassNotFoundException) instead of the expected Error (UnsupportedClassVersionError), which is not caught by the current catch clause, leading to a failure in <clinit>.

Request:
We kindly request that this pull request be expedited in an upcoming release. Its inclusion would enable us to upgrade to the latest version of the library, address CVE-2025-66021, and maintain compatibility with our environment.

Environment Details:
Application Server: WebLogic 12
Java Version: 1.8
Current Library Version: 20220608.1
Target Library Version: Latest version that addresses CVE-2025-66021 and includes the pull request fix #343

Thank you for your attention to this matter. Please let us know if further details are needed.
-- Jason

jmanico:
I merged the relevant PR as a step towards this being released for you.

jguyea:
> I merged the relevant PR as a step towards this being released for you.

Thank you Jim!

jmanico:
Fixed on `main` by #379 from @aleruz-dt: the `java8-shim` and `java10-shim` classes are now bundled inside `owasp-java-html-sanitizer.jar` and the shim artifacts are no longer published separately, so there is no split package on the module path and nothing extra to put on the classpath. If you added an explicit dependency on either shim as a workaround, remove it when you upgrade. This ships in the next release after 20260313.1. Closing; please reopen if the new release does not fix your case.

Issue #329, opened 2024-03-27
ClassNotFoundException: org.owasp.shim.Java8Shim after update to 20240325.1


veita:
After update to 20240325.1 there occur exceptions `ClassNotFoundException: org.owasp.shim.Java8Shim`.

I could not find documentation about which additional libraries have to be included on which Java platform such that things work properly. It seems that both `java8-shim` and `java10-shim` libraries are needed with Java 21, which seems odd.

Java version: OpenJDK Runtime Environment Temurin-21.0.2+13

randomalgorithm01:
@rombert @mikesamuel 

Can you kindly explain the use of this dependencies 8shim and 10shim. I originally thought the library has code that will compile in 11+ and this dependency act as adapter for older jdk. But from above it seems to be a problem in 21 jdk too.

I tried new jar in a large project without classnotfound exceptions but in a new scratch project build fails without shim jars? (Both in openjdk11)

Can you kindly explain need of this jar and when to add as dependency

randomalgorithm01:
@rombert a reply would be helpful herr

mebigfatguy:
Would be nice if these classes were loaded dynamically rather than statically. Why would i need to supply Java8-shim if i am not running on java8?

jlczuk:
This issue has been open for some time now.  I am hitting this as well.  All I did was update my version to 2024-325.1.

theigl:
Requiring both shims on the classpath makes the library unusable with the Java module system because of split-packages. Compilation fails with the following error:

`module XYZ reads package org.owasp.shim from both java10.shim and java8.shim`

aleruz-dt:
is there any news related to this issue? It applies also to the latest version, and as said by @theigl this is making the library unusable with the module system.

jmanico:
cc some of the new leads @mrabhishek @aalmiray @kittylyst @erikcostlow @brianf

aalmiray:
Understood. A viable modular option should be put in place. 

Please bear with us, as part of the new set of maintainers are getting used to the codebase.

aleruz-dt:
@aalmiray thanks for your effort.

aleruz-dt:
Hi @aalmiray — I opened PR #379 which directly addresses the JPMS split-package error described in this thread.

[...]

PR #379, opened 2026-04-23
Fix JPMS split-package error by bundling shims into the main JAR

### LINKED ISSUE #341: Please build the Java8/10 shim classes into the sanitizer JAR
The latest release removes the Guava dependency and introduces the use of Java8 and Java10 shims.  The sanitzer JAR was previously completely self-contained except for the Guava dependency.  It would be very helpful if the build of the JAR could include the shim classes rather than making them external dependencies.

Currently, I am required to identify the licenses and other information related to any JAR dependencies.  The current setup will no require me to get approval to use the java8-shim/java10-shim JARs as if they were separate dependencies from the sanitizer; the same thing I had to do for Guava previously.

aleruz-dt:
## Problem

Since release `20240325.1`, the library ships three separate JARs on Maven Central:

- `owasp-java-html-sanitizer`
- `java8-shim`
- `java10-shim`

Both `java8-shim` and `java10-shim` declare the package `org.owasp.shim`. When all three JARs are placed on the **JPMS module path** (Java 9+), the JVM raises a hard error at startup — before any application code runs:

```
Error occurred during initialization of boot layer
java.lang.LayerInstantiationException: Package org.owasp.shim in both
  com.googlecode.owasp-java-html-sanitizer.java8.shim
  and com.googlecode.owasp-java-html-sanitizer.java10.shim
```

This is a **split-package** violation. JPMS forbids the same package from appearing in more than one named or automatic module, and there is no workaround available to consumers short of keeping the library off the module path entirely.

The issue was first reported in #341 (June 2024). This PR resolves it.

---

## Root Cause

`Java8Shim` uses reflection at runtime to select the right implementation:

```java
// Prefers ForJava9AndLater (from java10-shim); falls back to ForJava8 (from java8-shim)
_instance = Class.forName("org.owasp.shim.ForJava9AndLater").newInstance();
```

This dispatch works correctly on the classpath. The problem is purely structural: JPMS resolves modules — and enforces split-package constraints — at layer instantiation time, before any reflection can execute. Two JARs exporting the same package is unconditionally rejected.

---

## Fix

Use `maven-shade-plugin` in `owasp-java-html-sanitizer` to **inline both shim JARs into the main artifact** at build time.

With `org.owasp.shim` living inside a single JAR, JPMS sees one module, one package — no split. The existing reflection-based dispatch in `Java8Shim` continues to work unchanged.

Both `java8-shim` and `java10-shim` remain published on Maven Central as standalone artifacts for backwards compatibility. They are marked `<optional>true</optional>` in the main module's POM so they no longer appear as transitive dependencies of consumers (their classes are now inlined).

An explicit `Automatic-Module-Name: owasp.java.html.sanitizer` is added to the JAR manifest so that the module name is stable and independent of the JAR filename, which is the recommended practice for libraries that have not yet migrated to a full `module-info.java`.

---

## Changes

| File | Change |
|---|---|
| `pom.xml` | Add `maven-shade-plugin 3.6.0` to `<pluginManagement>` (required by `requirePluginVersions` enforcer) |
| `owasp-java-html-sanitizer/pom.xml` | Add shade execution; mark shim deps `<optional>`; add `Automatic-Module-Name` to bundle manifest |
| `docs/maven.md` | Update stale version example; note that the JAR is now self-contained |
| `change_log.md` | Add entry for next release |

---

## Compatibility

**No source or binary incompatibility for the public API** (`org.owasp.html`). The following points are worth noting for the release notes:

- **Transitive dependencies**: `java8-shim` and `java10-shim` no longer appear in the transitive dependency graph of `owasp-java-html-sanitizer`. Projects that inadvertently relied on them as transitives must now declare them explicitly. In practice this is unlikely, as `org.owasp.shim` is an internal implementation package.
- **OSGi**: `org.owasp.shim` is not listed in `Export-Package` and was never part of the public OSGi contract. OSGi consumers are unaffected.
- **`banDuplicateClasses` enforcer rule**: this rule is currently commented out in the parent POM with a `TODO`. The shading intentionally causes `org.owasp.shim` classes to appear in both the shim JARs and the inlined copy inside the main JAR within the reactor build. This is consistent with the existing TODO and does not affect published artifacts.

---

## Verification

```bash
./mvnw verify
```

All tests pass. The OSGi manifest verification (`maven-verifier-plugin`) confirms `Export-Package: org.owasp.html` is present and correct in the shaded JAR.

Closes #341

### DIFF
--- a/owasp-java-html-sanitizer/pom.xml
+++ b/owasp-java-html-sanitizer/pom.xml
@@ -67,6 +67,10 @@
         <configuration>
           <instructions>
             <Export-Package>org.owasp.html</Export-Package>
+            <!-- Explicit JPMS automatic module name so consumers using the
+                 module path can require this module by a stable name that
+                 is independent of the JAR filename. -->
+            <Automatic-Module-Name>owasp.java.html.sanitizer</Automatic-Module-Name>
           </instructions>
         </configuration>
       </plugin>
@@ -86,6 +90,44 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals><goal>shade</goal></goals>
+            <configuration>
+              <!-- Inline java8-shim and java10-shim into this JAR so that
+                   consumers using the JPMS module path do not encounter a
+                   split-package error from org.owasp.shim appearing in
+                   multiple JARs. See https://github.com/OWASP/java-html-sanitizer/issues/341 -->
+              <artifactSet>
+                <includes>
+                  <include>com.googlecode.owasp-java-html-sanitizer:java8-shim</include>
+                  <include>com.googlecode.owasp-java-html-sanitizer:java10-shim</include>
+                </includes>
+              </artifactSet>
+              <filters>
+                <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
+                    <!-- Exclude shim Maven metadata to avoid duplicate
+                         pom.properties / pom.xml entries in the JAR -->
+                    <exclude>META-INF/maven/com.googlecode.owasp-java-html-sanitizer/java*/**</exclude>
+                  </excludes>
+                </filter>
+              </filters>
+              <!-- Replace the original JAR with the shaded one as the
+                   main artifact; do not attach a separate shaded artifact -->
+              <shadedArtifactAttached>false</shadedArtifactAttached>
+              <!-- Generate a reduced POM that omits the now-inlined shim
+                   dependencies, keeping the published POM accurate -->
+              <createDependencyReducedPom>true</createDependencyReducedPom>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 
@@ -93,10 +135,12 @@
     <dependency>
       <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
       <artifactId>java8-shim</artifactId>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
       <artifactId>java10-shim</artifactId>
+      <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>commons-codec</groupId>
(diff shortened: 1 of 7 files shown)


aleruz-dt:
@aalmiray can you have a look?

owasp-java-html-sanitizer/pom.xml
```diff
@@ -67,6 +67,10 @@
         <configuration>
           <instructions>
             <Export-Package>org.owasp.html</Export-Package>
+            <!-- Explicit JPMS automatic module name so consumers using the
+                 module path can require this module by a stable name that
+                 is independent of the JAR filename. -->
+            <Automatic-Module-Name>owasp.java.html.sanitizer</Automatic-Module-Name>
```
aalmiray, line 73: It would make more sense to use a proper module descriptor at this point. The module descriptor should be placed in the versioned space of the JAR, i.e, `META-INF/versions/9`

owasp-java-html-sanitizer/pom.xml
```diff
@@ -86,17 +90,57 @@
...
+        </executions>
+      </plugin>
     </plugins>
   </build>
 
   <dependencies>
     <dependency>
       <groupId>com.googlecode.owasp-java-html-sanitizer</groupId>
       <artifactId>java8-shim</artifactId>
+      <optional>true</optional>
```
aalmiray, line 138: If the shims are included in the final JAR then there's no sense in publishing them separately to Maven Central. They should be skipped from deployment.

aalmiray (changes requested):

jmanico:
Decision: we are going with this approach. It fixes the most-reported problem in the tracker (#329, #341, #376) and the module descriptor question can be handled separately.

I pushed one commit to this branch (thanks for leaving maintainer edits on, @aleruz-dt) that addresses @aalmiray's second point: `java8-shim` and `java10-shim` now set `maven.deploy.skip`, so once bundled they are no longer published on their own. Publishing them separately would only tempt people to add them back and recreate the split package. The change log entry now tells people to remove any explicit shim dependency they added as a workaround. I also dropped a stray `PLAN.md` line from `.gitignore` and bumped the version in `docs/maven.md`.

@aalmiray's first point, a real `module-info.class` in `META-INF/versions/9`, is tracked in #389 as a follow-up. @aalmiray, if you see a problem with merging the shade approach first, please say so here; otherwise this goes in once the PR build is green.

PR #428, opened 2026-09-08
Let a caller build on a CssSchema, and stop globally() widening a chosen one

### LINKED ISSUE #380: OWASP Java HTML Sanitizer: gaps and inconsistencies (v20260102.1)
The following six issues were discovered while integrating `owasp-java-html-sanitizer` into a
production HTML sanitization pipeline for user-authored rich content. All findings were verified
empirically against the published jar using reflection probes and live `sanitize()` roundtrips.
They range from missing CSS primitives (`fr` unit, `conic-gradient`) to API contract gaps that
make it impossible to override DEFAULT schema entries without reflection or full schema rebuilds.

## Environment

- owasp-java-html-sanitizer: `20260102.1`
- Java: 21
- Context: sanitizing user-authored HTML/CSS in form entries (rich content fields)

---

### 1. `CssTokens.UNIT_TRIE` missing `fr` unit

`fr` (the CSS Grid fractional unit, standardized in CSS Grid Layout Level 1 since 2017) is not
recognized by `CssTokens.isWellKnownUnit`. The `UNIT_TRIE` static initializer covers:

```
em, ex, ch, rem, vh, vw, vmin, vmax, px, mm, cm, in, pt, pc,
deg, rad, grad, turn, s, ms, hz, khz, dpi, dpcm, dppx
```

`fr` is absent. As a result, tokens like `1fr` lex as `BAD_DIMENSION` and the schema rejects
them, silently dropping entire grid track definitions.

**Reproduction:**

```java
PolicyFactory p = new HtmlPolicyBuilder()
    .allowElements("div").allowAttributes("style").globally()
    .allowStyling()
    .toFactory();

String input = "<div style=\"display:grid;grid-template-columns:1fr 280px\">x</div>";
System.out.println(p.sanitize(input));
// Actual:   <div style="display:grid;grid-template-columns:280px">x</div>
// Expected: <div style="display:grid;grid-template-columns:1fr 280px">x</div>
```

`1fr` is dropped; `280px` survives because `px` is a known unit. The grid collapses to a
single-column layout with no sanitizer warning.

**Suggested fix:** Add `fr` to the `UNIT_TRIE` static initializer in `CssTokens.java`. It is a
length-like quantity (non-negative integer or decimal, no negative values in practice) and
semantically analogous to `%`.

---

### 2. `background-image` DEFAULT literals omit `inherit`

The DEFAULT `CssSchema` entry for `background-image` has literals `{",", "none"}`. The keyword
`inherit` is missing. Every sibling property that also accepts a URL or image value includes
`inherit` in its literal set:

| Property            | `inherit` in literals? |
|---------------------|------------------------|
| `background-color`  | yes                    |
| `color`             | yes                    |
| `display`           | yes                    |
| `list-style-image`  | yes                    |
| `cursor`            | yes                    |
| **`background-image`** | **no**              |
| `content`           | no                     |

**Reproduction:**

```java
PolicyFactory p = new HtmlPolicyBuilder()
    .allowElements("div").allowAttributes("style").globally()
    .allowStyling()
    .toFactory();

// Stripped — background-image has no "inherit" literal:
System.out.println(p.sanitize("<div style=\"background-image:inherit\">x</div>"));
// Actual:   <div>x</div>
// Expected: <div style="background-image:inherit">x</div>

// Survives — list-style-image does have it:
System.out.println(p.sanitize("<div style=\"list-style-image:inherit\">x</div>"));
// Actual:   <div style="list-style-image:inherit">x</div>
```

**Suggested fix:** Add `"inherit"` to `background-image`'s literal set in
`CssSchema.DEFINITIONS`. This is a one-line change consistent with the pattern already
established by neighboring properties.

---

### 3. `allowUrlsInStyles(...)` silently no-ops without `allowUrlProtocols(...)`

Calling `allowUrlsInStyles(lambda)` registers a `styleUrlPolicy` on the builder. When
`compilePolicies()` assembles the `StylingPolicy`, the URL rewriter chains the lambda with a
`FilterUrlByProtocolAttributePolicy` built from the builder's global `allowedProtocols` set. If
`allowUrlProtocols(...)` was never called at the builder level, that set is empty, and the
protocol filter rejects every URL — including URLs the lambda explicitly approved.

The result is that `allowUrlsInStyles(IDENTITY_ATTRIBUTE_POLICY)` appears to be a no-op when
`allowUrlProtocols` is absent. There is no warning, no exception, and no diagnostic output.

**Reproduction:**

```java
PolicyFactory p = new HtmlPolicyBuilder()
    .allowElements("div").allowAttributes("style").globally()
    .allowStyling()
    .allowUrlsInStyles(AttributePolicy.IDENTITY_ATTRIBUTE_POLICY)
    // Without .allowUrlProtocols("https"), the line above has no effect.
    .toFactory();

String input = "<div style=\"background:url('https://example.com/img.png')\">x</div>";
System.out.println(p.sanitize(input));
// Actual:   <div>x</div>
// Expected: <div style="background:url('https://example.com/img.png')">x</div>

// Adding .allowUrlProtocols("https") makes it work as expected.
```

The interaction is documented nowhere in the Javadoc for `allowUrlsInStyles`.

**Suggested fixes (any one of):**

1. Add a Javadoc note to `allowUrlsInStyles` that it requires at least one `allowUrlProtocols`
   call to take effect.
2. Throw `IllegalStateException` at `toFactory()` time if `allowUrlsInStyles` was called but
   `allowedProtocols` is empty.
3. When the caller supplies an explicit lambda to `allowUrlsInStyles`, skip the protocol-filter
   chain and treat the lambda as the sole arbiter of URL acceptance.

---

### 4. `CssSchema.union()` cannot extend a DEFAULT property

`CssSchema.union(CssSchema a, CssSchema b)` throws `"Duplicate irreconcilable definitions"` when
both schemas define the same property key with differing bits, literals, or fn sets. This makes
it impossible to override a DEFAULT property without rebuilding the entire schema from scratch via
`withProperties()`.

Common use cases that hit this wall:

- Adding `inherit` to `background-image`'s literals (as described in issue #2 above).
- Adding `cursor:url()` support to the default `cursor` entry.
- Narrowing an overly permissive DEFAULT property for a specific deployment.

**Reproduction:**

```java
// Attempt to add "inherit" to background-image's literals:
CssSchema.Property backgroundImageWithInherit = CssSchema.properties(
    "background-image", "none,inherit"
).get("background-image");

Map<String, CssSchema.Property> overrides = new HashMap<>();
overrides.put("background-image", backgroundImageWithInherit);

// This throws: "Duplicate irreconcilable definitions for background-image"
CssSchema extended = CssSchema.union(CssSchema.DEFAULT, CssSchema.withProperties(overrides));
```

There is no `unionWithOverride` variant and no merge-strategy parameter.

**Suggested fix:** Add `CssSchema.unionWithOverride(CssSchema base, CssSchema overrides)` that
applies `overrides` entries on top of `base` entries when a conflict is detected, rather than
throwing. Alternatively, accept a `MergeStrategy` enum (`THROW`, `PREFER_FIRST`, `PREFER_SECOND`,
`MERGE_LITERALS`).

---

### 5. `CssSchema.withProperties(Map)` requires fn self-containment

`withProperties(Map<String, Property>)` throws `"Property map is not self contained"` if any
`Property`'s `fnKeys` references a key not also present in the input map. This means a single
custom property that uses CSS functions (e.g., a `background-image` Property that allows
`linear-gradient(`) cannot be built in isolation — the input map must also contain an entry keyed
`"linear-gradient("` for every function the property references.

Combined with the `union()` constraint in issue #4, there is no clean API path to:

1. Take the DEFAULT schema as a base.
2. Override a single property that uses functions.
3. Add new function entries alongside it.

Every attempt either throws on the `union()` call or requires the caller to reconstruct the
complete reachable closure of every property and function key.

**Reproduction:**

```java
// Attempting a minimal override: background-image that allows linear-gradient.
// This throws "Property map is not self contained" because "linear-gradient(" is not
// a key in the map, even though it exists in CssSchema.DEFAULT.
CssSchema.withProperties(Map.of(
    "background-image", /* property with fnKey "linear-gradient(" */
));
```

**Suggested fix:** Relax the self-containment check to allow fn keys that are already present in
a referenced base schema, or document the exact workaround (supply a complete map of all
properties reachable transitively via fn keys, mirroring whatever DEFAULT already provides).

---

### 6. Modern CSS function gaps

The following CSS functions are valid per current specifications but are not present in the
DEFAULT allowed-function list:

| Function           | Specification             | Typical use           |
|--------------------|---------------------------|-----------------------|
| `conic-gradient()` | CSS Images Level 4 (2019) | Backgrounds           |
| `repeat()`         | CSS Grid Layout Level 1   | `grid-template-*`     |
| `minmax()`         | CSS Grid Layout Level 1   | `grid-template-*`     |

`linear-gradient` and `radial-gradient` are present in DEFAULT. Their `conic-gradient`
counterpart is not, making the gradient support asymmetric. `repeat()` and `minmax()` are the
two primary value-building functions for CSS Grid tracks; without them, authors cannot express
responsive grid templates even after the `fr` unit gap (issue #1) is addressed.

**Reproduction:**

```java
PolicyFactory p = new HtmlPolicyBuilder()
    .allowElements("div").allowAttributes("style").globally()
    .allowStyling()
    .toFactory();

// conic-gradient stripped:
System.out.println(p.sanitize("<div style=\"background:conic-gradient(red, blue)\">x</div>"));
// Actual:   <div>x</div>
// Expected: <div style="background:conic-gradient(red, blue)">x</div>

// repeat() stripped — grid template collapses:
System.out.println(p.sanitize(
    "<div style=\"grid-template-columns:repeat(3, 1fr)\">x</div>"));
// Actual:   <div>x</div>
// Expected: <div style="grid-template-columns:repeat(3, 1fr)">x</div>
```

**Suggested fix:** Add `conic-gradient(`, `repeat(`, and `minmax(` to the function definitions in
`CssSchema.DEFINITIONS` alongside `linear-gradient(` and `radial-gradient(`.

---

## Why these matter

Modern CSS is increasingly common in user-authored content, particularly in rich-text editors and
form-based authoring tools. CSS Grid (`fr`, `repeat()`, `minmax()`) is now baseline browser
support and appears regularly in authored HTML. `conic-gradient` has been widely supported since
2021. `inherit` on `background-image` is valid CSS 2.1 behavior.

The API ergonomics issues (#3, #4, #5) compound the data gaps (#1, #2, #6): when downstream
users discover a missing value or unit, the natural response is to extend the DEFAULT schema. The
current API makes that extension path either silent-failing (#3) or structurally blocked (#4, #5).
Addressing the ergonomics issues would allow the ecosystem to self-correct without waiting for
upstream releases.

jmanico:
Two commits: #380 items 4 and 5, and a security fix found while doing them.

## 1. Let a caller build on a CssSchema (#380 items 4, 5)

Extending what DEFAULT allows for one property wasn't just awkward, it was **impossible without reflection**, and for a reason the issue didn't identify: `Property`'s constructor is public and takes a bitfield, but `bits`, `literals` and `fnKeys` had no accessors and `forKey` was package-private. There was no way to *read* an existing definition to base a new one on. `union()` then refused the result, since two schemas defining a property differently is exactly what it throws on.

**`Property`** gains `bits()`, `literals()`, `fnKeys()`, plus `withLiterals`, `withoutLiterals` and `withFunctions` to derive a modified copy without hand-copying the other two fields. The `BIT_*` constants become public — documenting what the public constructor already required callers to supply — and there's a `toString` so failures stop printing identity hashes.

**`CssSchema`** gains `property(name)` and `withOverrides(map)`:

```java
CssSchema schema = CssSchema.DEFAULT.withOverrides(
    Collections.singletonMap(
        "cursor", CssSchema.DEFAULT.property("cursor").withLiterals("zoom-in")));
```

Overrides resolve function keys against the **combined** schema, so a property may refer to a function the base already defines without supplying it again — that's item 5. A key neither side defines is still rejected, since it would otherwise resolve to `DISALLOWED` silently at runtime.

`union` keeps throwing on conflicts. Silently picking a winner would hand a caller a policy they didn't write; `withOverrides` is how you say which one wins, and `union`'s javadoc now points there. The self-containment message names the offending property and the way out.

**This widens nothing on its own** — a caller could already build any schema via `withProperties(Map)`; they just had to reconstruct all 136 DEFAULT entries to change one.

*Note on item 5:* I kept `withProperties(Map)` throwing for a standalone schema. There the function really is unreachable, so failing loudly beats silently disallowing it. The message now points at `withOverrides` for the build-on-a-base case.

## 2. Do not let allowing style globally widen a chosen schema

Found while testing the above: an overridden schema passed to `allowStyling()` threw `Duplicate irreconcilable definitions` instead of being used. The cause is worse than the symptom.

`allowAttributes("style").globally()` called `allowStyling()`, which installs `CssSchema.DEFAULT`, and `allowStyling(schema)` unioned onto whatever was there. So a caller who named a narrower schema **silently got DEFAULT back**:

```java
.allowAttributes("style").globally()
.allowStyling(CssSchema.withProperties(Arrays.asList("color")))

input:  color:red;font-weight:bold;width:10px
before: color:red;font-weight:bold;width:10px    <- all of DEFAULT
after:  color:red
```

No error, no warning — the policy was just more permissive than the one written down. Reversing the calls didn't help; the union ran either way.

`globally()` still installs a default so styling is sanitized rather than passed through, but only when the caller hasn't chosen a schema, and it marks that schema implicit. An explicit `allowStyling()` replaces an implicit schema rather than unioning with it. **Two explicit calls still union**, the documented behaviour, now covered by a test.

This tightens policies rather than loosening them, so the risk of the change is that someone was unknowingly relying on the extra permissiveness.

## Verification

Full matrix locally — JDK 11, 17, 21, 25, `mvn verify`, **413 tests** (up from 404), exit 0 on all four.

Closes #380.

https://claude.ai/code/session_01BZfawfTUAFHN7cocUjcpQK

### DIFF
--- a/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java
+++ b/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java
@@ -65,15 +65,26 @@ public final class CssSchema {
 
     /**
      * @param bits A bitfield of BIT_* constants describing groups of allowed tokens.
-     * @param literals Specific allowed values.
+     * @param literals Specific allowed values.  Converted to lower case, since
+     *     values are matched against a lower-cased token; a literal that was
+     *     not lower case could never match.
      * @param fnKeys Maps lower-case function tokens to the schema key for their parameters.
+     *     The function tokens are converted to lower case for the same reason.
      */
     public Property(
         int bits, Set<String> literals,
         Map<String, String> fnKeys) {
       this.bits = bits;
-      this.literals = j8().setCopyOf(literals);
-      this.fnKeys = j8().mapCopyOf(fnKeys);
+      Set<String> literalsBuilder = new HashSet<>();
+      for (String literal : literals) {
+        literalsBuilder.add(Strings.toLowerCase(literal));
+      }
+      this.literals = j8().setCopyOf(literalsBuilder);
+      Map<String, String> fnKeysBuilder = new HashMap<>();
+      for (Map.Entry<String, String> e : fnKeys.entrySet()) {
+        fnKeysBuilder.put(Strings.toLowerCase(e.getKey()), e.getValue());
+      }
+      this.fnKeys = j8().mapCopyOf(fnKeysBuilder);
     }
 
     /**
@@ -102,7 +113,7 @@ public final class CssSchema {
      * <p>This widens what the property accepts.  Prefer naming the exact
      * values you need over reaching for a broader {@code BIT_*} bit.
      *
-     * @param extraLiterals lower-case values to add.
+     * @param extraLiterals values to add; converted to lower case.
      * @return a new property; this one is unchanged.
      */
     public Property withLiterals(String... extraLiterals) {
@@ -117,12 +128,14 @@ public final class CssSchema {
      * <p>Values not currently allowed are ignored, so this is safe to use
      * against a schema whose exact contents you have not pinned down.
      *
-     * @param unwantedLiterals lower-case values to remove.
+     * @param unwantedLiterals values to remove; matched case-insensitively.
      * @return a new property; this one is unchanged.
      */
     public Property withoutLiterals(String... unwantedLiterals) {
       Set<String> narrowed = new HashSet<>(literals);
-      narrowed.removeAll(Arrays.asList(unwantedLiterals));
+      for (String unwanted : unwantedLiterals) {
+        narrowed.remove(Strings.toLowerCase(unwanted));
+      }
       return new Property(bits, narrowed, fnKeys);
     }
 
--- a/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java
+++ b/owasp-java-html-sanitizer/src/main/java/org/owasp/html/CssSchema.java
@@ -27,6 +27,7 @@
 
 package org.owasp.html;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -75,6 +76,79 @@ public final class CssSchema {
       this.fnKeys = j8().mapCopyOf(fnKeys);
     }
 
+    /**
+     * A bitfield of {@code BIT_*} constants describing groups of allowed
+     * tokens.
+     */
+    public int bits() { return bits; }
+
+    /**
+     * The specific values this property allows, as an immutable set.
+     */
+    public Set<String> literals() { return literals; }
+
+    /**
+     * Maps lower-case function tokens such as {@code "rgb("} to the schema key
+     * describing their arguments, as an immutable map.
+     *
+     * <p>A key named here must also be present in the schema, or the function
+     * resolves to {@link CssSchema#DISALLOWED} and its values are dropped.
+     */
+    public Map<String, String> fnKeys() { return fnKeys; }
+
+    /**
+     * This property, also allowing the given values.
+     *
+     * <p>This widens what the property accepts.  Prefer naming the exact
+     * values you need over reaching for a broader {@code BIT_*} bit.
+     *
+     * @param extraLiterals lower-case values to add.
+     * @return a new property; this one is unchanged.
+     */
+    public Property withLiterals(String... extraLiterals) {
+      Set<String> widened = new HashSet<>(literals);
+      widened.addAll(Arrays.asList(extraLiterals));
+      return new Property(bits, widened, fnKeys);
+    }
+
+    /**
+     * This property, no longer allowing the given values.
+     *
+     * <p>Values not currently allowed are ignored, so this is safe to use
+     * against a schema whose exact contents you have not pinned down.
+     *
+     * @param unwantedLiterals lower-case values to remove.
+     * @return a new property; this one is unchanged.
+     */
+    public Property withoutLiterals(String... unwantedLiterals) {
+      Set<String> narrowed = new HashSet<>(literals);
+      narrowed.removeAll(Arrays.asList(unwantedLiterals));
+      return new Property(bits, narrowed, fnKeys);
+    }
+
+    /**
+     * This property, also allowing the given functions.
+     *
+     * <p>The schema must also define the keys these map to, or the functions
+     * resolve to {@link CssSchema#DISALLOWED}.  {@link CssSchema#withOverrides}
+     * checks that for you.
+     *
+     * @param extraFnKeys maps a lower-case function token such as
+     *     {@code "rgb("} to the schema key for its arguments.
+     * @return a new property; this one is unchanged.
+     */
+    public Property withFunctions(Map<String, String> extraFnKeys) {
+      Map<String, String> widened = new HashMap<>(fnKeys);
+      widened.putAll(extraFnKeys);
+      return new Property(bits, literals, widened);
+    }
+
+    @Override
+    public String toString() {
+      return "[CSS property bits=" + bits + " literals=" + new TreeSet<>(literals)
+          + " fns=" + new TreeSet<>(fnKeys.keySet()) + "]";
+    }
+
     @Override
     public int hashCode() {
       final int prime = 31;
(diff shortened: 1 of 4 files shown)


jmanico:
Added a third commit from a review pass over this branch and #427. Two real defects, both the same shape as the `globally()` bug above — a caller believing a policy is narrower than it is.

**1. Property literals were not canonicalized.** Literals are matched against a lower-cased token, so a literal that is not itself lower case can never match. Nothing enforced that:

```java
new CssSchema.Property(0, Collections.singleton("ZigZag"), ...)   // allows nothing
CssSchema.DEFAULT.property("color").withoutLiterals("RED")        // removes nothing
```

The second is the dangerous direction: a caller narrowing a schema silently doesn't. And it lands on hand-built properties, which is exactly the workflow this PR is meant to enable. The constructor now lower-cases literals and function tokens, so the derivation helpers inherit it. Every built-in definition is already lower case, so nothing changes for them — the existing suite is unaffected.

**2. #427's guard test had a hole.** `testDefaultGrantsNoLayoutControl` asserts over the layout family, but missed nine properties that #427 itself had defined: `grid-auto-columns`, `grid-auto-rows`, the four `grid-{row,column}-{start,end}` placements, and the three `grid-gap` aliases. All nine are correctly absent from DEFAULT — I verified — but the guard wasn't pinning them, which is the entire point of a guard. Now covered.

Also documented on `allowStyling(CssSchema)` that naming a schema replaces the implicit default and that two explicit calls union, since that's the rule the second commit changes.

Matrix re-run: JDK 11/17/21/25, **414 tests**, exit 0 on all four.

PR #478, opened 2026-09-11
Rework the literal-text tag filter for fidelity and footprint

### LINKED ISSUE #470: The literal-content tag filter mangles script text with comparison operators
Follow-up from the review of #465, which reworked the filter that strips tags from the text of a kept `script`, `style` or `iframe` element.

The filter models a tag as `<`, optional whitespace, a letter, and everything up to the next `>`, the way the CVE-2025-66021 fix did. Ordinary JavaScript trips it. With `allowElements("script").allowTextIn("script")`:

| input | output |
|---|---|
| `if (a < b) { x(); } if (c > d) { y(); }` | `if (a  d) { y(); }` |
| `for(i=0;i<n;i++){a[i]=b>c;}` | `for(i=0;ic;}` |

`< b` is taken as a tag opening (a browser reads `<` followed by whitespace as text), and `<n;i++){a[i]=b>` is a tag named `n;i++){a[i]=b`, which a browser would also read as a tag in a markup context but which is inert inside script text. Before #465 the damage ran to the end of the chunk; now the tag alone goes, so this is better but still lossy.

The content is literal to every browser, so none of this is a security matter; it is fidelity for policies that let script or style text through. Options, each with a trade-off:

- Follow the tokenizer and stop treating `<` plus whitespace as a tag start. Cheap, and browser-faithful, but test #6 of the CVE-2025-66021 fixtures deliberately strips `< script>` inside style text.
- Only strip what could matter: end tags, which are the breakout vector, and leave start tags as text. Simplest and most faithful, but tests #2 and #3 of those fixtures expect an injected `<script>` or `<div>` inside a style block to go, content and all.
- Recognise only well-formed tag names, letters, digits and hyphens, so `<n;i++)...>` stays text. Keeps the current shape and fixes the second row but not the first.

Whichever is chosen should keep the chunk-boundary rule from #465: a `<` at the end of a chunk, or followed by `/`, must still go.

### LINKED ISSUE #473: The literal-content tag filter keeps a record per tag and can exhaust the heap on large inputs
Follow-up from the review of #465.

`stripTags` in `ElementAndAttributePolicyBasedSanitizerPolicy` keeps a record per tag for the whole chunk, an `int[4]` plus an `ArrayList` slot, a boxed `Integer` and an `ArrayDeque` entry, roughly 45 to 75 bytes per tag of attacker-controlled input, where the filter before #465 kept only the output buffer. With `allowElements("style").allowTextIn("style")` and an input of `<style>` plus three million `<b>x` plus `</style>` (12 MB), the current code throws `OutOfMemoryError` at `-Xmx160m` and needs about 256 MB, while the previous code sanitized the same input at 160 MB. Distinct tag names cost more.

The records exist only to pair each start tag with its matching end tag so that a dropped element's content goes with it, which buys no safety: whatever survives is tag-free literal text either way. Two directions:

- Drop the pairing: one linear pass that deletes every recognised tag and keeps the text between. Simpler, allocation-free, no quadratic cliff to guard. Fidelity trade: `<style>a{}<script>alert(1)</script>c{}</style>` becomes `a{}alert(1)c{}`, inert style text, and fixture tests #2 and #3 of the CVE-2025-66021 set expect the content to go too.
- Keep the pairing but store records in flat `int` arrays with a per-name chain of previous unmatched starts, roughly 16 bytes per tag, or the package's `IntVector`.

Either way, `tagNameOf` should stop using `String.split("\\s")`, which compiles a `Pattern` per tag and is about half of the filter's time on the 200,000-tag test; scanning for the first HTML whitespace does the same job.

### LINKED ISSUE #474: The literal-content gate ignores the renderer's foreign-content escaping and custom receivers
Follow-up from the review of #465.

`isLiteralContentElement` in `ElementAndAttributePolicyBasedSanitizerPolicy` re-derives the renderer's escaping decision from `HtmlTextEscapingMode` through `HtmlStreamRenderer.safeName`, but the renderer has a second input it does not: inside `svg` and `math` a `style` or `script` element's text is escaped as RCDATA, because browsers parse it as markup there. So a kept `<style>` inside `<svg>` now has its tags stripped by the policy where the renderer alone escaped and preserved them. With `allowElements("svg","style","b").allowTextIn("style","svg")`:

```
<svg><style>a{}<b>x</b>c{}</style></svg>
```

gave `<svg><style>a{}&lt;b&gt;x&lt;/b&gt;c{}</style></svg>` before #465 and gives `<svg><style>a{}c{}</style></svg>` now. Safe either way, but lossy for no reason.

The related edge: through the `safeName` indirection the policy leaves `xmp`, `listing` and `plaintext` text unstripped because the shipped renderer renames them to `pre` and escapes; a custom `HtmlStreamEventReceiver` handed to `PolicyFactory.apply` does not rename, so it receives `</noscript>` as a text event verbatim. The renderer-side check added in #472 does not run for such a receiver either.

One shared predicate that both the policy and the renderer consult, taking the foreign-content depth into account, would keep the two from drifting.

### LINKED ISSUE #475: Deleting a tag from literal text can splice a comment delimiter and drop the whole element content
Follow-up from the review of #465.

When `stripTags` deletes a tag, or a matched start-to-end region, from kept `style` or `script` text, the text on either side is spliced together and can form a `-->`, or lose the `-->` that closed an earlier `<!--`. `HtmlStreamRenderer.checkHtmlCdataCloseable` then rejects the whole buffered body, so the entire content of the element is dropped where the previous filter kept it. With `allowElements("style","script","b").allowTextIn("style","script")`:

| input | before #465 | now |
|---|---|---|
| `<style>a{}-<b>->b{}</style>` | unchanged | `<style></style>` |
| `<style>a{}--<b>>b{}</style>` | unchanged | `<style></style>` |
| `<script><!-- if (a > 0) { <b> } --> </b> f();</script>` | unchanged | `<script></script>` |

In the last row the matched `<b> --> </b>` pair goes and takes the comment's `-->` with it, leaving `<!--` unclosed. Fail-safe, since the content is dropped rather than emitted, and reported through `discardedText` since #463, but lossy. Only fragments beginning with `-` can join this way, because a `<` before a deleted tag is taken along with it. Replacing a deleted tag with a single space would prevent the joins; checking the body once at close time instead of per chunk would remove the class.

jmanico:
Closes #470, closes #473, closes #474, closes #475.

## Summary

The four follow-ups from the review of #465 all live in the same ~120 lines, so
the filter on the text of a kept literal-content element is rewritten once, as a
single pass that keeps no record per tag. Built on #477, which had just landed.

**#470 — a tag needs a well-formed name.** The filter took `<`, optional
whitespace, a letter and everything to the next `>` for a tag, so ordinary
script and style text lost the text between its brackets. A name is now an ASCII
letter and then letters, digits, `-`, `_`, `:` or `.`, ending at whitespace, `/`
or `>`; running into anything else the output keeps makes it no name:

| input | before | now |
|---|---|---|
| `if (a < b) { x(); } if (c > d) { y(); }` | `if (a  d) { y(); }` | unchanged |
| `for(i=0;i<n;i++){a[i]=b>c;}` | `for(i=0;ic;}` | unchanged |
| `var s = 'a<b'; var t = 'c>d';` | `var s = 'ad';` | unchanged |

Fixture #6 still strips `< script>`, and a name the renderer's elision would
otherwise assemble (`</noscript` NUL `>`, DEL, a C1 control) is still removed —
`Encoding` gains `isPossiblyElidedCodeunit` for that, since a policy has to
judge the text that will be emitted.

I checked the relaxation against Chrome 152 rather than the spec alone: a weird
name **is** a live element with a live handler where a browser parses markup
(`<b)` → `localName "b)"`, `onmouseover` a function, fires), so the question is
whether such text can reach a markup context. It cannot: an end tag that would
break a raw-text element open has a well-formed name and still goes, the
renderer refuses content holding one anyway (#472), text in a literal element
under a `select` is dropped by the balancer (CVE-2021-42575), and in foreign
content the renderer escapes. Chrome also confirms `</b)>` does not end a
raw-text `noscript`, while `</noscript/>`, `</noscript >` and `</NOSCRIPT>` do
— all of which the filter removes.

**#473 — the filter's memory is its output.** No `int[]` per tag, no `ArrayList`,
no `ArrayDeque`, no boxed `Integer`, and `tagNameOf`'s `String.split("\\s")` is
gone. Pairing reads a bounded note of the start tags not yet matched and
compares names where they lie; removed ranges are recorded only while a listener
is attached. Fixtures #2 and #3 pin start-tag-plus-content removal, so the
pairing stays — bounded, because it buys fidelity rather than safety.

- `<style>` + 3M × `<b>x` + `</style>` (12 MB): `OutOfMemoryError` at `-Xmx160m`
  before, sanitizes at `-Xmx96m` now.
- Allocation for 200k tags: 836 bytes/tag before, 111 now — of which 104 is what
  the lexer spends on the same text either way, so the filter's own share went
  from ~730 bytes/tag to ~7. The new test measures that difference.

**#475 — a removal must not splice something new.** Joining the text on either
side of a removed tag could make a `<!--` or `-->`, and a dropped pair could
take a `-->` with it; either left the element's comments unbalanced, which costs
its whole content:

| input | before | now |
|---|---|---|
| `<style>a{}-<b>->b{}</style>` | `<style></style>` | `<style>a{}- ->b{}</style>` |
| `<style>a{}--<b>>b{}</style>` | `<style></style>` | `<style>a{}-- >b{}</style>` |
| `<script><!-- if (a > 0) { <b> } --> </b> f();</script>` | `<script></script>` | content kept |

A space goes in only where the join would make a delimiter, so no existing
expectation moves; the element's text is remembered across chunk boundaries,
because the lexer splits literal text at `<%...%>` and the join can straddle
chunks (`<style>a{}-<%%><b>->b{}</style>` was empty too). A pair that would take
a delimiter with it is dropped tag by tag instead.

**#474 — one predicate for where text is emitted as written.** `HtmlStreamRenderer`
now has the only copy of that decision and the policy asks it, with the
foreign-content depth the policy tracks alongside its other per-element gates:

- `<svg><style>a{}<b>x</b>c{}</style></svg>` keeps its tags again, escaped by the
  renderer as RCDATA, instead of having them stripped for nothing.
- A receiver handed to `PolicyFactory.apply` does not rename `xmp`, `listing` and
  `plaintext` to `pre`, so it used to receive their text with a `</noscript>`
  intact, and the renderer-side check does not run for it. With any receiver but
  the library's own renderer, every element whose content the lexer read as raw
  text is filtered.

`HtmlChangeReporter.OutputChannel` becomes one of the library's own decorators so
the renderer stays visible behind it; otherwise attaching a listener would change
what the filter strips, which a new test pins.

## Two more holes the fuzzing turned up

Both are on `main` today and are fixed here, with the splice tests beside the
#475 ones:

- `<style><b<svg> onmouseover=alert(1)>x</style>` emitted
  `<b onmouseover=alert(1)>x` inside the style body: the input's tag was
  `<b<svg>` and what followed it was text, so removing the tag turned that text
  into a live handler. A sweep now takes the `<` of anything a removal, or the
  element's own end tag, would finish into a tag, while `<3` and `< b` stay.
- `<style></<b>noscript></style>` handed a receiver `</noscript>` verbatim — a
  real breakout for a custom receiver, caught only by the renderer's backstop on
  the `sanitize` path. It now emits `/noscript>`.

## Test plan

- 18 new or reworked tests in `HtmlSanitizerTest`; 8 of them fail against
  `main`'s library code, one per issue and then some, and the listener test fails
  without the reporter change. 551 tests pass in all.
- Three linearity tests guard the scan's worst cases, each of which I measured
  quadratic first: 1M `<` that open no tag with one `>` at the end (101 s → 92
  ms), a 1.5M-long run of unfinished tags ended by a removable one (12 s → 80
  ms), and the existing 200k unmatched tags.
- Local fuzzing beyond the suite: 2.4M hostile literal-text cases, delivered
  whole and one character at a time, with an oracle for an end tag a browser
  acts on and for a start tag a browser acts on. `main` leaves ~3,900 tag shapes
  per 200k cases, including the `</noscript>` above; this leaves none. Output
  idempotence also improves (14.2% → 10.1% of cases differ on a second pass; the
  remainder is #465's dangling-`<` rule, unchanged here).
- `./mvnw clean verify` passes on JDK 11, 17, 21 and 25, fuzzers, AntiSamy suite
  and JPMS consumer check included.

## Fidelity left on the table

`if (a<b || c>d) f();` still loses ` b || c`: a name followed by whitespace is a
tag wherever a browser reads markup, exactly like the `< script>` that fixture #6
pins, and the filter cannot tell them apart. A test records that.

https://claude.ai/code/session_01L4kNbd4REPF5ozRjnCu6tW

### DIFF
--- a/change_log.md
+++ b/change_log.md
@@ -2,6 +2,49 @@
 
 Most recent at top.
   * Next release
+    * Ordinary script and style text survives the filter on kept
+      literal-content elements.  A tag now needs a well-formed name -- an
+      ASCII letter and then letters, digits, `-`, `_`, `:` or `.` -- so
+      `if (a < b) { x(); } if (c > d) { y(); }` and
+      `for(i=0;i<n;i++){a[i]=b>c;}` keep the text between the brackets that
+      used to be read as a tag and removed.  Anything a browser would read as
+      a tag only where it parses markup stays, which the text of a kept
+      literal-content element never reaches; a name that a character the
+      renderer elides would otherwise hide, as in `</noscript` NUL `>`, is
+      still recognised and removed.  Issue #470.
+    * The filter keeps no record per tag, so a chunk of literal text full of
+      tags no longer costs a multiple of its own size in heap: input that used
+      to sanitize in 160 MB and needed 256 MB after the filter was reworked
+      now needs less than it ever did.  A start tag is paired with its end tag
+      from a bounded note of the start tags not yet matched, names are
+      compared where they lie instead of being copied and split with a regular
+      expression, and the ranges removed are recorded only while a listener is
+      attached.  Issue #473.
+    * Removing a tag from literal text no longer leaves what is on either side
+      of it able to open a tag.  `<style><b<svg> onmouseover=1>` emitted
+      `<b onmouseover=1>`, turning text that followed a tag into a tag with a
+      live handler, and `<style></<b>noscript>` emitted `</noscript>`, which
+      the renderer then refused but which reached a receiver handed to
+      `PolicyFactory.apply` as written.  The `<` of anything a removal, or the
+      element's own end tag, would finish into a tag now goes, while a `<` that
+      opens no tag, as in `<3` or `< b`, stays.
+    * Removing a tag from literal text no longer joins the text on either side
+      of it into a `<!--` or `-->` that the input did not hold, and a start
+      tag dropped together with its content no longer takes an enclosed `-->`
+      with it.  Either one left the element's comment delimiters unbalanced,
+      which cost the whole content: `<style>a{}-<b>->b{}</style>` and
+      `<script><!-- if (a > 0) { <b> } --> </b> f();</script>` came out empty.
+      Issue #475.
+    * The filter now runs exactly where the receiver writes text as it stands.
+      Inside `svg` and `math` a browser parses the content of every element as
+      markup, and `HtmlStreamRenderer` escapes it there, so the tags in the
+      text of a `style` or `script` element in foreign content are kept and
+      escaped rather than stripped.  A receiver handed to
+      `PolicyFactory.apply` does not rename `xmp`, `listing` and `plaintext`
+      to `pre` as the renderer does, and so used to receive the text of those
+      elements with its tags intact, a `</noscript>` among them; the text of
+      every element the lexer reads as raw text is now filtered for any
+      receiver but the library's own renderer.  Issue #474.
     * The filter on text kept inside `style`, `script`, `iframe` and other
       literal-content elements now examines a possible tag prefix before
       looking for its closing `>`.  A long run of `<` characters before one
--- a/owasp-java-html-sanitizer/src/main/java/org/owasp/html/ElementAndAttributePolicyBasedSanitizerPolicy.java
+++ b/owasp-java-html-sanitizer/src/main/java/org/owasp/html/ElementAndAttributePolicyBasedSanitizerPolicy.java
@@ -27,11 +27,9 @@
 
 package org.owasp.html;
 
-import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.BitSet;
-import java.util.Deque;
-import java.util.HashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
@@ -99,6 +97,21 @@ class ElementAndAttributePolicyBasedSanitizerPolicy
    */
   private @Nullable String keptCdataElementName;
   /**
+   * True while a kept {@code svg} or {@code math} element is open, so that a
+   * browser parses the content of every element inside it as markup.
+   * Maintained like {@link #skipText}, and in step with the renderer's own
+   * count of the same elements, since the renderer sees the tags this policy
+   * emits.  {@link #isLiteralContentElement} says why it matters.
+   */
+  private boolean inForeignContent;
+  /**
+   * The last few characters emitted for the kept literal-content element that
+   * is open.  Text arrives in chunks, and {@link #stripTags} needs them to see
+   * a comment delimiter that removing a tag would otherwise splice together
+   * across a chunk boundary.
+   */
+  private String literalTextTail = "";
+  /**
    * Alternating input names and adjusted names of elements opened by the
    * caller.
    */
@@ -111,8 +124,16 @@ class ElementAndAttributePolicyBasedSanitizerPolicy
   private final BitSet skipTextBeforeOpen = new BitSet();
   /** The same for {@link #inKeptCdataElement}. */
   private final BitSet inKeptCdataBeforeOpen = new BitSet();
+  /** The same for {@link #inForeignContent}. */
+  private final BitSet inForeignContentBeforeOpen = new BitSet();
   /** The same for {@link #keptCdataElementName}; entries may be null. */
   private final List<String> keptCdataNameBeforeOpen = new ArrayList<>();
+  /**
+   * True if {@link #out} is the library's own renderer, possibly behind
+   * {@link HtmlStreamEventReceiverWrapper} decorators, so that what it escapes
+   * is known.  See {@link #isLiteralContentElement}.
+   */
+  private final boolean outIsLibraryRenderer;
 
   /** Told about filtered literal content; null while nobody is listening. */
   private @Nullable HtmlStreamRenderer.DroppedTextListener droppedTextListener;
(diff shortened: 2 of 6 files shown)


jmanico:
**Claude Code review** — 1 finding(s)

**nit** — owasp-java-html-sanitizer/src/main/java/org/owasp/html/HtmlStreamRenderer.java:66
Stale Javadoc {@ link #foreignContentRootElementNames} references a field renamed in this diff to FOREIGN_CONTENT_ROOT_ELEMENT_NAMES



---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #485, opened 2026-09-11
The element at the nesting limit is written but its end tag is not


jmanico:
Found while fixing #342 (#481), which does not change it.

`TagBalancingHtmlStreamEventReceiver` writes a start tag when the stack has fewer than `nestingLimit` entries, so the element at index `nestingLimit - 1` is written. But the close paths gate on `openElements.size() < nestingLimit` before the pop, in `prepareForContent`, and on `last + 1 < nestingLimit` in `closeTag`, which is one stricter, so that element's end tag is never written when it is closed by an end tag or pushed out by content it cannot hold. The element after it is then written inside it, and the output is unbalanced:

```java
balancer.setNestingLimit(3);
// <div><div><span>x</span><p>y</p></div></div>
// now: <div><div><span>x<p>y</div></div>
```

`closeDocument` uses `Math.min(nestingLimit, size)` and so gets it right for whatever is still open at the end. The policy and renderer downstream see the same unbalanced events, and a browser closes the elements itself, so the sanitizer's own model stays consistent with the browser; but the balancer's stack no longer matches the output from that point, and elements the author closed stay open in the output. The default limit is 256, so this needs input nested that deep. The fix is to gate every close on the index the element was written at, `index < nestingLimit`, as #481 does for the tags it adds. Reproduced on `main` at fad01c1.

jmanico:
One more piece of the same picture: the implied-elements step in `prepareForContent` writes and pushes implied elements without checking the limit at all, so a stack can hold entries at index `nestingLimit` and beyond that were written, while the resume loop pushes an entry it did not write once the stack is at the limit. So "written" cannot be read off the index alone at present. A fix should make the three paths agree: gate the write on `size < nestingLimit` everywhere, and gate every close on the same index rule.

Issue #483, opened 2026-09-11
A table part arriving outside a table context implies a new table where a browser returns to the open one or drops the part


jmanico:
Found while fixing #342 (#481), which leaves these unchanged.

`TagBalancingHtmlStreamEventReceiver.prepareForContent` opens implied elements from the current top of the stack before it checks what the top can contain, so a table part arriving in a container that is not a table gets a new table implied there, even when a browser would return to a table already open or ignore the part.

1. A row arriving inside a cell's `div` starts a new table inside the `div`; a browser closes the cell and adds the row to the open table:

```
<table><tr><td><div>x<tr><td>y</td></tr></table>
now:     <table><tbody><tr><td><div>x<table><tbody><tr><td>y</td></tr></tbody></table></div></td></tr></tbody></table>
browser: <table><tbody><tr><td><div>x</div></td></tr><tr><td>y</td></tr></tbody></table>
```

2. A row arriving inside a `p` with no table open gets a table implied inside the `p`, and a browser does not keep a table inside a `p`, so the output re-parses differently and is not idempotent:

```
<div><p>x<tr><td>y</td></tr>
now:     <div><p>x<table><tbody><tr><td>y</td></tr></tbody></table></p></div>
browser reads the output as: <div><p>x</p><table>...</table><p></p></div>
```

3. A row arriving after a bare `col` gets a table implied inside the `colgroup`, again not idempotent:

```
<table><col><tr><td>y</td></tr></table>
now:     <table><colgroup><col /><table><tbody><tr><td>y</td></tr></tbody></table></colgroup></table>
browser: <table><colgroup><col></colgroup><tbody><tr><td>y</td></tr></tbody></table>
```

The common cause is that the implied-elements step does not ask whether the element it implies can go inside the current top, and does not look for an open table to return to; a browser's "clear the stack back to a table context" does both. #481 added that return for tables it pushed out of the way, and the same walk, guarded by the table-scope boundaries (`template`, and foreign content once #461 lands), could serve here. Cases 2 and 3 are stability bugs, since the output is read back as a different tree; they are layout rather than safety, since nothing changes parsing context. Reproduced on `main` at fad01c1.

jmanico:
Related, found while testing #481: the balancer treats `</template>` like any other end tag and looks for the template within default scope, so a table implied inside a template blocks it:

```
<template><tr><td>y</td></tr></template>z
now: <template><table><tbody><tr><td>y</td></tr></tbody></table>z</template>
[...]

Issue #482, opened 2026-09-11
Text inside an element nested in a rename into a literal-content element bypasses the target's text gate


jmanico:
Found while reviewing #469 (fixed in #480), which now reports the drop but does not change it.

`ElementAndAttributePolicyBasedSanitizerPolicy` keeps an open-element entry for every element it emits, and gates text by the nearest kept entry. When an `ElementPolicy` renames an element into one whose content `HtmlStreamRenderer` writes literally, such as `style`, the renderer refuses every tag that arrives inside it, but the policy does not know that: it pushes the nested element as kept, opens the text gate for it, and emits its text, which the renderer appends to the literal content.

```java
PolicyFactory f = new HtmlPolicyBuilder()
    .allowElements((name, attrs) -> "style", "div")
    .allowElements("b", "style")   // no allowTextIn("style")
    .toFactory();
f.sanitize("<div>a<b>bold</b>c</div>");
// <style>bold</style>
```

`a` and `c` are dropped, because text was not allowed in `style`, but `bold` reaches the stylesheet as raw text through the gate the policy opened for `b`. The closeability check in the renderer still refuses `</style` and comment tricks in that text, so it stays inside the element, and the only route to it is a policy that renames an element into a raw-text one, so this is a policy-author mistake rather than a bypass. But text lands where the author said no text may go, and the policy's stack and the renderer's disagree about what is open for the rest of that element.

Suggested fix: while a kept literal-content element is open, the policy should treat a nested start tag as dropped (`deferOpenTag`), whether or not text is allowed in the literal element, so that the text inside it follows the literal element's own gate and the reporter sees the drop as the policy's. That needs a flag for "inside a kept literal-content element" separate from the current one, which is set only when text is allowed there. Then the renderer's literal-content drop site would be unreachable through the library policy and remain as defense for other callers.

Reproduced on `main` at fad01c1.

goutamadwant:
I opened #489 to address this.

It keeps nested start tags from becoming separate text containers when an
`ElementPolicy` renames the outer element to literal content. Text now follows
the outer element's gate, while `allowTextIn` continues to preserve explicitly
allowed text. The change also reports the nested tag as the policy's discard.

The full `./mvnw -ntp -B clean verify` build passes with the new positive and
negative regressions.

Issue #342, opened 2024-06-20
Issue with HTML Sanitization: Improper Handling of <div> Tag Inside <table>


sumitkumar1110:
Hi,
We are using the this library in  [Zimbra](https://github.com/Zimbra/java-html-sanitizer-release-20190610.1) to sanitize customer-generated HTML content in emails. During this process, we encountered an issue where a <div> tag inside a <table> tag causes improper sanitization. Specifically:

1.  The sanitizer closes the <table> tag before the <div> tag and reopens it after the <div>.
2. The <div> tag does not close where it originally should; instead, it closes just before the end of the HTML document.
It seems that the sanitizer uses a stack to manage tags and the <div> tag remains in the stack until all other tags are processed, causing it to close at the end of the document.

Before Sanitization
`<table align="center" border="0" cellpadding="0" cellspacing="0" class="full-width-mobile" role="presentation" style="width:500px;background:#fff;" width="500">
     <div class="mobilecontent mobilecontent-gmail" style="mso-hide:all;display:none;max-height:0px;overflow:hidden;padding:0;height:0;"> 
      
      <tr style="padding-top:0; padding-bottom:0;">
       <td class="show-mobile-table-cell gmail-td-hide" style="padding-top:0;padding-bottom:0;display:none;mso-hide:all;"><table class="show-mobile-table gmail-table-hide" align="center" border="0" cellpadding="0" cellspacing="0" role="presentation" style="width:100%;display:none;mso-hide:all;">
         <tr>
          <td class="show-mobile-table-cell pad-horiz-mobile gmail-td" colspan="2" style="display:none;mso-hide:all;background:#eee;color:#444;padding-top:20px;padding-bottom:20px;font-family:Roboto, Arial, sans-serif; font-size:20px; -mso-line-height-rule:exactly;line-height:24px;font-weight:500;width:100%">Ważna uwaga dotycząca usługi: wsparcie dla sprzętu wygasło </td>
         </tr>
         <tr>
          <td class="show-mobile-table-cell" valign="bottom" style="vertical-align:bottom;border-bottom: 1px solid #01447C;padding-bottom:10px;padding-top:45px;padding-left:15px;display:none;mso-hide:all;">
            <a href="https://urllink/path/sfasdfas" target="_blank"><img src="https://images.com/LOGO_A.jpg" alt="new logo" width="171" style="width:171px; display: block;" border="0"></a>
      </td>
[...]

PR #490, opened 2026-09-18
Balance forms in table modes and close foreign elements by identity (#484)

### LINKED ISSUE #484: A form directly inside a table stays open as a container
Found while fixing #342 (#481), which leaves this unchanged.

A browser inserts a `form` start tag that arrives directly inside a table and pops it at once, so nothing that follows is inside the form. The tag balancer keeps the `form` open as an ordinary container, so the text and rows after it nest inside it, and a browser reads the output as a different tree:

```
<table><form>x<tr><td>y</td></tr></table>
now:     <table><form>x<table><tbody><tr><td>y</td></tr></tbody></table></form></table>
browser reads the output as: x<table><form></form></table><table><tbody><tr><td>y</td></tr></tbody></table>
browser reads the input as:  x<table><form></form><tbody><tr><td>y</td></tr></tbody></table>
```

The element tables say a table can contain a `form`, which is true of the tree a browser builds, but the browser's "form element pointer" handling means the form has no content there. A fix would treat a `form` opened directly in a table, row group or row as closed immediately, the way #481 treats a void element. Layout rather than safety. Reproduced on `main` at fad01c1.

jmanico:
> **Review status: fixes complete, ready for your review.** Thirteen commits on top of
> `19b5fd2`, ending at `33c2b90`.  The second review's fifteen findings were each
> checked against `main`, the base and both heads: the two that held up are fixed in
> `33c2b90`, `f3f4406`'s content loss in `f2fc154`.  What remains is filed as #491
> (follow-up documentation) and #492 (pre-existing on main). The review of `19b5fd2` is in
> [this comment](https://github.com/OWASP/java-html-sanitizer/pull/490#issuecomment-5730317234).
> Eight blocking findings were reproduced against `main` @ `569d463`; the event-balance win
> (96,766 broken streams fixed, 0 regressed) holds up, and no XSS path was found.
>
> Blockers, in suggested order of work:
>
> - [x] 1. `TagBalancing:3400` `indexOfPassthroughNamed` is O(N^2) on every end tag —
>       `"<foo></div>" x 32000` takes 5.2 s vs main's 8.4 ms (remote DoS)
> - [x] 2. `TagBalancing:635` a `<form>` dropped at the nesting limit latches the form
>       pointer, so depth >= 256 emits **zero** forms for the rest of the document
> - [x] 3. `TagBalancing:681` prepared-vs-raw element name mismatch throws
>       `IllegalStateException` out of `sanitize()` behind a renaming `withPreprocessor`
> - [x] 4. `HtmlChangeReporter:215` attaching an `HtmlChangeListener` changes the sanitized
>       bytes behind a `withPostprocessor`
> - [x] 5. `TagBalancing:3907` / `:3911` allowed text is silently deleted
>       (`<tbody><form>B`, `<ul><caption>E`, `<ul><svg><tbody>D`)
> - [x] 6. `TagBalancing:645` synthetic `<li>` around foreign roots — `<select><svg>` grows
>       +9 chars on every re-sanitize pass, unbounded
> - [x] 7. `TagBalancing:1012` in-table `<form>` reverses document order and splits one table
>       into two
> - [x] 8. `HtmlSanitizerTest:2334` and `:2519` pin finding 7 as expected and are the only
>       tests in their group omitting the sibling `parseAsBrowser` assertion — add it first
>
> Six further findings are listed in the comment as leads that were not independently
> reproduced.
>
> The fixes, the verification of them, and what was left are in
> [this comment](https://github.com/OWASP/java-html-sanitizer/pull/490#issuecomment-5733077353).
> The performance follow-up, two small items and a correction are in
> [this comment](https://github.com/OWASP/java-html-sanitizer/pull/490#issuecomment-5733530152).
> The blocking regression, its fix and a correction to the second review are in
> [this comment](https://github.com/OWASP/java-html-sanitizer/pull/490#issuecomment-5734246154).
> The third pass, with every finding classified against `main`, is in
> [this comment](https://github.com/OWASP/java-html-sanitizer/pull/490#issuecomment-5736375250).

> **Third pass (review of `33c2b90`, comment
> [5738700026](https://github.com/OWASP/java-html-sanitizer/pull/490#issuecomment-5738700026)):
> seven more commits, ending at `6c1fb9f`.** Fixed and verified per finding; the fix series and
> its verification are in the comment that follows it.
>
> - [x] 1. A dropped `select` wraps its contents and what follows in a stray `<li>` under
>       `Sanitizers.BLOCKS` — `eb11921`
> - [x] 2. Children of a dropped or renamed foreign root beside a table deleted — `c772698`
> - [x] 3. Cell text of a dropped table inside an integration point deleted — `cf8bad6`
> - [x] 4. Formatting resumed inside the list implied for a list item — `3387e8c`
> - [x] 5. (found during acceptance) A run of unclosed unknown tags followed by ordinary
>       markup was quadratic, 11.7 s at 128k+128k tags — `cfb21b6`
> - [x] 6. (independent pass) `disallowTextIn("form")` bypassed by a suppressed cell in an
>       integration point — `bd1bf23`
> - [x] 7. (independent pass) Formatting resumed after its own end tag, a link after the next
>       link — `6c1fb9f`

Fixes #484. Follow-up to #481 (#342), #486 (#485), #487 (#483) and #489 (#482).

## Summary

A `form` start tag that arrives directly in a table, row group or row is inserted and popped at
once by a browser, so nothing after it is inside the form. The tag balancer kept it open as an
ordinary container. This branch makes the balancer follow the browser's form-element-pointer
rules, and while reviewing that change we found and fixed a class of stack-bookkeeping bugs in
how SVG/MathML and unrecognized end tags are matched to open elements.

Commits, oldest first:

1. `14e4d29` Handle forms in table insertion modes. The original #484 change: a form in table
   context is written as an empty element; the input pointer stays set until a real `</form>`.
2. `69cde7e` Track form pointer, foreign and template contexts for #484 review fixes. Recovered
   work from earlier review sessions: input and output form-pointer tracking, foreign-namespace
   bits for stack entries, exact foreign contexts inside retained templates, policy-produced
   output names, nesting-limit forwarding, and fixes for review findings A (recursive implied
   wrapper) and C (stale outputless wrapper).
3. `6fd3a1f` Close popped foreign elements by identity and balance forwarded elements. Replaces
   the provisional name-based fix for finding B and resolves finding D (five counterexamples):
   - every node tracked by `ForeignContentContext` gets an opaque serial; the balancer stores it
     with the stack entry for the same start tag, so a foreign end tag closes exactly the entries
     of the nodes it popped, innermost first, and never an older element of the same local name;
   - elements forwarded without a stack entry (svg, math, custom elements, table parts forwarded
     directly into foreign output) are recorded with the stack depth at which they were
     forwarded; their end tags are forwarded only while they are open below, after everything
     opened inside them is closed, so the policy's input-name stack pops one entry per close;
   - a stray end tag closes nothing, a root dropped at the nesting limit gets no close, a table
     part clears such elements back to a real table context, formatting closed this way resumes
     as a browser reconstructs it, and `closeDocument` closes them in LIFO order;
   - removes the policy-report reconciliation, the ownership flags, and the
     `openElements.size() < nestingLimit` guard for unrecognized end tags that this replaces.
4. `e92ffb3` Forget forwarded elements inside a suppressed option by count, not depth. A
   precision fix in the policy-only suppressed-option path.
5. `bceeb7f` Do not resume formatting where content is inserted under foreign rules. Fixes the
   one defect the independent review found in commit 3: formatting closed with an integration
   point's end tag was resumed inside SVG or MathML output, where a browser never reconstructs
   formatting and where `b` or `i` pops the foreign root. Formatting now resumes only where the
   next tag or text is inserted under HTML rules, and the input tracker follows a forwarded
   element the balancer closed by name when a browser would have ignored that end tag.
6. `19b5fd2` Judge formatting resumption by the insertion point, not by the tag. The review's second pass
   showed commit 5 still resumed formatting in front of a tag that itself breaks out of foreign
   content (`p`, `hr`, `table`), inside the foreign root. The insertion point is now judged
   before the tag, so the formatting resumes for the content inside the breakout element, and
   a void breakout adds no empty formatting element. Plain HTML behavior is unchanged.

## Behavior changes visible to callers

- `TagBalancingHtmlStreamEventReceiver` used directly over a plain receiver now closes
  unrecognized elements at `closeDocument` and before an enclosing element closes, and no
  longer forwards an end tag with no matching start. One existing test expectation
  (`HtmlSanitizerTest.testSelfClosingTagEvents`) gained the trailing `closeTag path` that this
  produces; the self-closing behavior it tests is unchanged.
- Formatting elements closed by an unrecognized end tag (`<foo><b>x</foo>y`) now resume around
  later content, matching the browser's active-formatting reconstruction.
- No public API changes. New methods are on package-private types only.

## Validation

- `./mvnw clean verify` on JDK 11, 17, 21 and 25 at the final commit: BUILD SUCCESS on each, 653
  library tests and 7 example tests, JPMS consumer check included.
- Fuzzers at fixed seeds 1, 484 and 20250912 on JDK 21 at the final commit: green.
- Focused hostile regressions added for findings A, B, C and D.1 to D.5, on the policy, the
  change-reporter wrapper and a plain receiver, with exact idempotence, validator.nu browser-tree
  and strict event-balance checks; positive cases included.
- Differential probe (`Issue484Differential`, 5000 random cases per seed, 14 policies, plus
  event cases at nesting limits 0 to 8) against `main` (`569d463`), the pre-#481 commit
  (`fad01c1`) and the pre-repair recovered state, compared per case key:
  - versus `main`: 136,917 sanitize cases and 47,688 event cases that failed there (non-idempotent,
    browser tree mutates on reparse, or unbalanced events) are clean now; 123 sanitize and 169
    event cases that were clean there now report a problem. Every sanitize case in that set is
    under a drop or rename policy, and the browser tree of the output stays stable across the
    second pass; the event cases are plain-receiver runs at limits 0 to 8 whose balanced output
    is then not a fixed point of the keep policy, which the probe conflates with a failure.
  - versus the pre-repair recovered state: 350 sanitize and 43,072 event cases fixed (the event
    cases were unbalanced plain-receiver streams); 51 sanitize and 2 event cases newly report a
    problem. Replaying each first-pass output on the pre-repair build shows 39 of the 51 are
    idempotence gaps that build already has on that output, exposed because the foreign roots are
    now closed where a browser closes them. The remaining 12 are exact-serialization differences
    with stable browser trees: 9 where a dropped template promotes its contents and 3 for one
    input where a `</td>` inside `foreignObject` makes the tracker unknown on the first pass.
- Hostile shape set (nested tables, two pushed-out runs, integration points, mixed-case names,
  literal elements around pushed content, attribute-less table rename/drop policies): no case
  regressed versus either baseline; one improved to a browser-equal tree.
- Independent adversarial review by a fresh Opus session of commit 3 (about 190,000 fuzzed
  inputs with a strict LIFO checker between policy and renderer, plus targeted shapes): zero
  event-balance problems on either receiver; one High finding (fixed in commits 5 and 6; the
  review's second pass on commit 5 confirmed the fix, found no new balance or identity problem
  in another 186,000 fuzzed cases, and reported the breakout residue that commit 6 closes; my
  360-shape integration-point sweep on the final code shows no formatting emitted at a foreign
  insertion point and stable reparse trees); one Low finding kept as a documented limitation (a
  table part arriving inside a `select` now clears an unrecognized element before the implied
  table reopens, which can move that implied table outside the `select`); one Low idempotence
  case under a template-dropping policy with a stable browser tree; and an informational note
  that node serials wrap after 2^32 start tags.

## Known limitations (pre-existing, not addressed here)

- A plain receiver still gets legacy HTML table structure implied inside SVG and cannot retire
  foreign output on an HTML breakout start, so `svg, td, /svg` and `svg, noscript, b, /svg`
  event streams stay unbalanced as on `main`.
- Policies that drop or rename tables or templates are not closed under their own output;
  exact idempotence can fail on hostile garbage under such policies while the browser tree stays
  stable.
- The foreign tracker becomes unknown on a table-scoped end tag inside an integration point,
  after which end tags fall back to name-based matching.
- Formatting queued when an integration point closes resumes for the text inside a breakout
  element such as a table cell, where a browser's cell marker would stop reconstruction; the
  balancer's containment model does not track formatting markers there. The reparse tree is
  stable either way.
- The balancer does not foster-parent unrecognized elements out of tables, so `<table><foo>`
  keeps `foo` inside the table in the output, though it now closes it when a table part arrives.
- Node serials are 32-bit and wrap after about four billion tracked start tags in one document.

Do not merge until the review comments are resolved.


jmanico:
## Review of `19b5fd2` vs `main` @ `569d463` — please do not merge yet

Two passes: a max-effort `/code-review` (10 finder angles, then adversarial verification),
and an independent differential probe of my own. Everything below was measured by building
**both** worktrees and diffing their behaviour, on JDK 21 / macOS arm64, using the
nu.validator parser already on the test classpath as the browser oracle.

**Ten of the fifteen findings I reproduced myself and quote real output for.** The rest are
marked as reported-but-not-independently-reproduced, so please treat those as leads rather
than as established facts.

The PR is a large net win on its stated goal. The problem is collateral damage, and almost
all of it traces to one commit.

---

## 1. Quadratic time on attacker-controlled input — remote DoS

`TagBalancingHtmlStreamEventReceiver.java:3400`, `indexOfPassthroughNamed`.

It linearly scans the unbounded `passthroughNames` list, re-canonicalizing every entry, on
**every** end tag. `pushPassthrough` (3393) is uncapped: an unrecognized tag never enters
`openElements`, and when the policy drops it `writeOpenTag` is never reached, so
`outputNestingDepth` stays 0, `effectiveNestingDepth()` returns 0 forever, and the
`< nestingLimit` guard never fires.

Policy `Sanitizers.BLOCKS.and(FORMATTING).and(LINKS).and(TABLES)`, input `"<foo></div>"` x N:

| N | main | this branch |
|---|---|---|
| 2,000 | 3.1 ms | 25.2 ms |
| 4,000 | 2.9 ms | 77.5 ms |
| 8,000 | 5.4 ms | 300.6 ms |
| 16,000 | 5.4 ms | 1,213.2 ms |
| 32,000 | 8.4 ms | **5,152.2 ms** |

x4.0–4.25 per doubling is textbook O(N^2). A 352 KB request burns 5.2 CPU-seconds against
main's 8.4 ms (613x), and the output is empty in every case. It is every end tag, not only
unrecognized ones — the call site at 3326 runs for recognized ones too.

Suggested fix: store the canonical name at push time and index by it; count passthroughs in
`effectiveNestingDepth()`.

## 2. Silent deletion of allowed text — three independent paths

Not restructuring. Text the policy allows disappears from the output.

```
policy(table, form)
  <tbody><form>B            main <table><form>B</form></table>      HEAD <table><form></form></table>
  <tbody><form>B</form>C    main <table><form>B</form></table>C     HEAD <table><form></form></table>
  <td><form>B  <thead><form>B  <tr><form>B  <caption><form>B        same shape, B lost

policy(table, ul, li)
  <ul><caption>E            main <ul><li></li></ul>E                HEAD <ul><li></li></ul><table></table>
  <ul><td>E                 main <ul><li></li></ul>E                HEAD <ul><li></li></ul><table></table>

policy(select, svg, math, ul, li, tbody, option)
  <ul><svg><tbody>D</tbody></svg></ul>
                            main <ul><svg><tbody><li>D</li></tbody></svg></ul>
                            HEAD <ul><li><svg><tbody></tbody></svg></li></ul>       (D gone)
```

The browser tree of `<tbody><form>B` is `<form>"B"`; of `<ul><caption>E` is `<ul> "E"`.

`TagBalancingHtmlStreamEventReceiver.java:3907` (form path) and `:3911` (dropped-table-part
path) — the retired form or the implied table leaves the insertion point inside an emitted
`<table>`, and `text()`'s foster-parenting preparation never fires.

## 3. `IllegalStateException` escapes `sanitize()`

`TagBalancingHtmlStreamEventReceiver.java:681`. The policy start is prepared under
`canonElementName` but emitted under the raw `elementName`, so when they differ the prepared
state is never consumed and the next start tag throws.

Reproduced through the public API with a documented `HtmlPolicyBuilder.withPreprocessor(...)`
that renames `custom` to `CusTom`:

```
pf.sanitize("<custom><p>x</p></custom>")
  main  ->  <p>x</p>
  HEAD  ->  java.lang.IllegalStateException: A start tag is already prepared
            at ElementAndAttributePolicyBasedSanitizerPolicy:460 via TagBalancing:855
```

Before throwing, the mismatch runs `applyPolicies` **twice** on the same caller-owned attrs
list, so a stateful user `ElementPolicy`/`AttributePolicy` sees the tag twice and
`discardedAttributes` is computed from an already-filtered list.

Suggested fix: forward `canonElementName` at 801/803/807, and clear prepared state in
`openTag` whenever it exists but `usesPreparedStart` is false.

## 4. One deep input disables `<form>` for the rest of the document

`TagBalancingHtmlStreamEventReceiver.java:635`. `markFormElementPointerSet()` runs before the
nesting-limit check, so a `<form>` the limit then drops still latches the pointer; and the
`formElementPointerSet = false;` deleted from `HtmlSanitizer.becomeUnknown()`
(`HtmlSanitizer.java:1035`) lets the bit survive.

Policy allowing `form`, `div`; input `("<div>" x D) + "<form>a" + ("</div>" x D) + "<form>b</form>"`:

| D | main | HEAD |
|---|---|---|
| 255 | 2 `<form>` | 1 `<form>` |
| 256 | 1 `<form>` | **0 `<form>`** |
| 257 | 1 `<form>` | **0** |
| 300 | 1 `<form>` | **0** |

The nesting limit is documented as bounding depth, not as permanently disabling an element
type for the remainder of the document. Mark the pointer only after the form is accepted.

## 5. Attaching an `HtmlChangeListener` changes the sanitized bytes

`HtmlChangeReporter.java:215`. `InputChannel.outputElementNameForLastOpenTag()` answers from
its own `OutputChannel` observation instead of delegating to the wrapped policy, so behind a
`withPostprocessor` the balancer balances against the post-processor's filtered output.

Same policy, same input `<table><col><li><div></tr><textarea><foreignObject>`:

```
f.sanitize(in)                    <table><colgroup><col /><ul><li><textarea>&lt;foreignObject&gt;</textarea></li></ul></colgroup></table>
f.sanitize(in, listener, ctx)     <table><colgroup><col /><ul><li></li></ul></colgroup></table><textarea>&lt;foreignObject&gt;</textarea>
```

main is byte-identical both ways. An audit listener must never change what is sanitized.
Suggested fix: delegate to `((OpenTagOutputPolicy) policy).outputElementNameForLastOpenTag()`
when the wrapped policy is one.

## 6. Output grows on every re-sanitize pass

`TagBalancingHtmlStreamEventReceiver.java:645`. Widening the foreign-root guard from
`needsFosterParenting(elIndex)` to `!usesForeignContentRules` routes every HTML-inserted
`<svg>`/`<math>` through `prepareForContent` with `elIndex == UNRECOGNIZED_TAG`, so the legacy
containment tables imply a synthetic `<li>` no browser creates.

```
<select><svg>   pass1 (37) <select><li><svg></svg></li></select>
                pass2 (46) <select><ul><li><svg></svg></li></ul></select>
                pass3 (55) <select><li><ul><li><svg></svg></li></ul></li></select>
                pass4 (64) <select><ul><li><ul><li><svg></svg></li></ul></li></ul></select>
   main: stable at 28 chars, <select><svg></svg></select>
```

+9 chars per pass, unbounded. An application that re-sanitizes stored HTML on save grows the
row on every save. Same for `<select><math>`. Also `<ul><svg></svg></ul>` gains a synthetic
`<li>` that main and the browser do not produce.

The guard at 1960 that suppresses the synthetic `li` only fires when
`pushedOutTablePolicy() == null` (raw receiver) and only for `top == SELECT_TAG`, so list
containers are unprotected on every path.

## 7. In-table `<form>` reverses document order and splits the table

`TagBalancingHtmlStreamEventReceiver.java:1012`. The rule retires the output table, so content
a browser foster-parents **before** the table is serialized **after** it.

With this PR's own `HtmlSanitizerTest.formTablePolicy()`:

```
<table><form id=f><b>q</b></table>
  main  <table><form id="f"><b>q</b></form></table>
  HEAD  <table><form id="f"></form></table><b>q</b>
  browser tree of the input:  <b>"q"  THEN  <table><form>
```

main's output reparses to exactly the browser tree; HEAD's reparses with the `<b>` after the
table. With a tbody-less policy (`allowElements("table","tr","td","form")`, a very common
hand-written shape),
`<table><tr><td>a</td><form action=1><td>b</td></tr></table>` puts cells `a` and `b` in
**different tables**.

Root cause: `outputSuffixMatchesInput` (1454) used as a proxy for the output parser's
insertion mode.

## 8. The tests pin the regression, so CI cannot see it

`HtmlSanitizerTest.java:2334` `testTableFormDoesNotContainFollowingContent`, and
`:2519` `testDroppedTableModeForm`.

The three sibling tests added in the same hunk (`testFormStartIsPoppedInTableModes`,
`testFormsInCellsCaptionsAndTemplates`, `testTableFormsKeepBrowserFormPointerSemantics`) each
end every case with `assertEquals(parseAsBrowser(c[0]), parseAsBrowser(out), c[0])`. These two
omit it and assert only the literal string plus idempotence — and adding the sibling
assertion **fails**, because

```
parseAsBrowser(input) = ["x", <table><form id=f><tbody><tr><td>"y", "tail"]
parseAsBrowser(out)   = [<table><form id=f>, "x", <table>...]
```

12 `parseAsBrowser` calls exist across the 49 added tests; 38 of the added tests have none.
Adding the assertion to those two would turn finding 7 red today, which is the cheapest way to
keep it from regressing again. AGENTS.md: "Round-trip safety: sanitized output, re-parsed by a
browser, must not mutate into a different, unsafe DOM."

---

## Reported by the review pass, not independently reproduced by me

Please confirm these before acting on them.

- **`HtmlSanitizer.java:1035`** — `becomeUnknown()` keeping `formElementPointerSet` also makes
  the balancer apply the HTML form-pointer rule to `<form>` start tags that are really
  SVG/MathML elements. *(I did reproduce this one:* `<svg></tr><form><form>` *gives one form on
  HEAD, two on main and in the browser tree; and*
  `<svg></tr><form><b></b><form>x</form></form>` *merges the inner SVG form's text into the
  outer on HEAD.)*
- **`:356`** — `underlying instanceof FormPointerPolicy` / `PushedOutTablePolicy` feature
  detection went from 4 sites on main to 31 here. Because the interfaces are package-private,
  wrapping a policy in any third-party decorator silently switches it to a different balancing
  algorithm and no third-party `Policy` wrapper can opt in.
- **`:2098`** — `insertionPointIsInForeignContent` judged for the current tag rather than for
  where the queued formatting element lived; `<math><b></math><li>` reportedly emits the single
  input `<b>` twice and is not a fixed point.
- **`:837`** — implied `<tbody>`/`<tr>` emitted for a table part even when the policy drops
  `<table>`, leaving table parts with no enclosing table.
- **`:2096`** — formatting resumed *in front of* a breakout start tag (`p`, `hr`, `table`),
  placing it outside, where a browser ends up with the formatting inside. The commit message of
  `19b5fd2` states the opposite intent.
- **`:807`** — `pushPassthrough` runs even when the policy emitted nothing for the forwarded
  element, reportedly yielding a bare `<td>`/`<tr>`/`<option>` with no enclosing table or select.
- Smaller: a `<form>` dropped because the pointer is set is never reported to
  `HtmlChangeReporter` (630 returns before `underlying.openTag`), so audit listeners
  under-count removals; `closeStackEntriesFrom` (3506) resumes formatting out of nested
  template contents; `reopenedTableWasRenamed`
  (`ElementAndAttributePolicyBasedSanitizerPolicy.java:181`) is the one new mutable field never
  reset in `openDocument()`/`closeDocument()`; `setNestingLimit` (458) now mutates state before
  validating and can throw where main did not.

---

## What holds up

The headline claim is real and I confirmed it at scale. Differential probe, 34,000 hostile
inputs x 8 policies, plus the plain receiver at nesting limits 0–6:

- **Plain-receiver event balance: 96,766 broken streams fixed, 0 regressed.** main leaves
  13,188 unbalanced streams per 4,000-input run; this branch leaves none.
- Policy-to-renderer balance: 0 unbalanced on both.
- Idempotence 1,882 fixed / 161 exposed; browser-tree stability 1,827 fixed / 188 exposed. All
  161 exposures are the pre-existing implied-wrapper class — every one of the 129 distinct
  inputs contains `select`/`option`/`optgroup`/`li`/`col`, and
  `Sanitizers.BLOCKS.sanitize("<select>y")` already gives `<li>y</li>` -> `<ul><li>y</li></ul>`
  on **main**. Not caused by this PR, though it routes more inputs into it.
- **No allowlist widening**: every element name newly appearing in output is already allowed by
  that policy.
- 60,000 hostile foreign-content / raw-text cases (SVG and MathML integration points, mixed
  case, raw text, `annotation-xml`): **zero script elements, zero event-handler attributes,
  zero `javascript:` URLs** on either build. No XSS path found.
- `./mvnw clean verify` green on JDK 11, 21 and 25: 653 library + 7 example tests, JPMS
  consumer check included.
- Fuzzers green at five further fixed seeds beyond the three in the description: 7, 99, 1234,
  20260918, 424242.
- Public and protected API **byte-identical** to main (`javap -protected` over every public
  class) — the "no public API changes" claim checks out.
- Housekeeping clean: `change_log.md` bullet under `* Next release`, no tabs, no over-long
  added lines, **no test weakened** (both test files are +N/-0; `testSelfClosingTagEvents` only
  gained `"closeTag path"`, which tightens it).

## Where it comes from

Bisected with clean builds at each commit. Commit 2, `69cde7e` "Track form pointer, foreign and
template contexts for #484 review fixes", introduces the whole performance regression:

| shape, 8,000 repeats | `14e4d29` | `69cde7e` |
|---|---|---|
| well-formed table | 6.91 ms | 6.80 ms |
| unclosed `<table><tr><td>x` | 6.56 ms | 56.15 ms |
| unclosed `<table><form>w<tr><td>v` | 9.88 ms | 64.80 ms |
| unclosed `<b><i><a href=/x>q</zzz>` | 18.96 ms | 47.91 ms |

Well-formed input is unaffected; malformed input pays 4–10x, on top of the separate quadratic
blow-up in finding 1. Findings 3, 5, 7 and the `instanceof` proliferation also land in that
same commit.

The deeper cause is altitude: the balancer sits **upstream** of the policy
(`HtmlSanitizer.java:1345`), so most of the ~1,500 new lines exist to let it predict and
retroactively patch what the downstream policy will emit — the `outputElements` shadow stack,
six output-namespace BitSets, three form-pointer BitSets, the `passthrough*` third stack, the
speculative `prepareForStartTag` protocol, 31 `instanceof` checks, and 16 parallel
stack-index-keyed structures hand-synchronised at 14+ sites (two of which already disagree: the
pop loop at 2881 omits `pushedOut.clear(i)`, the one at 2916 omits
`outputlessTablesWithEmittedParts.clear(top)`). Running tree construction over the post-policy
stream would delete all of it.

## Suggested order of work

1. Findings 1 and 4 look contained — an index by canonical name plus counting passthroughs in
   `effectiveNestingDepth()`, and moving the pointer latch after acceptance.
2. Finding 3 is a two-line forwarding fix plus a prepared-state clear.
3. Finding 5 is a delegation fix in `InputChannel`.
4. Findings 2, 6 and 7 go to the design above and are the real work.
5. Add the sibling `parseAsBrowser` assertion to the two tests in finding 8 first, so 7 is red
   while it is being fixed.

Per AGENTS.md "small, single-purpose changes only, no drive-by refactors of parser, policy or
renderer code", it may be worth landing #484 on its own and taking the foreign-identity and
form-pointer rewrites as separate PRs with their own regression suites.

Nothing here needs a security advisory: this is an unmerged branch and none of it is an XSS
bypass.

jmanico:
## Review fixes: eight commits on top of `19b5fd2`

One commit per finding, each with regression tests that fail on `19b5fd2` for the stated
reason, in the order the review suggested. Nothing else in the branch was touched.

| # | Finding | Commit | Fix |
|---|---|---|---|
| 1 | Quadratic time on `"<foo></div>" x N` | `f3f4406` | Forwarded elements count toward the nesting limit, which bounds the passthrough list (and, for the first time, the depth a plain receiver sees for unrecognized elements); each entry keeps its canonical name so lookup no longer canonicalizes every entry. 32,000 repeats: **5,152 ms → 21 ms**, x1.4–1.8 per doubling. |
| 2 | Silent deletion of allowed text | `2c2d2bf` + `147f3fa` | When the policy closes a table's output ahead of fostered text, it restored the gate from *inside* the table, which never admits text. It now uses the gate before the table, as for any dropped element. All six `<…><form>B` shapes keep `B`; `<ul><caption>E` keeps `E`; so do `<td>E` and `<table><caption>E`, which lost their text on `main` too. |
| 3 | `IllegalStateException` out of `sanitize()` | `756d10c` | Unrecognized starts are forwarded under their canonical name, like recognized ones, so the prepared policy result is consumed. `<custom><p>x</p></custom>` through a recasing preprocessor → `<custom><p>x</p></custom>`. |
| 4 | Nesting-limit form latches the pointer | `58dd112` | The pointer is cleared again when the limit drops the form that set it. Depth 256/300 → the later `<form>b</form>` is emitted. Same commit: when the input tracker is unknown, the pointer rules follow the *output* namespace, so `<svg></tr><form><form>` keeps both SVG forms as `main` did. |
| 5 | Listener changes the bytes | `16e132d` | `InputChannel.outputElementNameForLastOpenTag()` delegates to the policy, falling back to the observed name only for a policy that cannot say. `sanitize(in)` and `sanitize(in, listener, ctx)` are byte-identical behind the postprocessor. |
| 6 | `<select><svg>` grows every pass | `350cc3a` | A foreign root under HTML rules is run through the containment metadata only in a table insertion mode (table, row group, row, column group — the colgroup recovery the widening was for is kept). `<ul><svg></svg></ul>` is unchanged; `<select><svg>` is stable at 28 chars over four passes. The policy-emitted-root path had the same widening and is gated the same way. |
| 7 | In-table form splits the table | `147f3fa` | The eager retirement fired on *any* output/input mismatch. A dropped `tbody`/`thead`/`tfoot`/`tr` is repaired identically by the output parser, so it no longer counts; a dropped template still does. `<table><tr><td>a</td><form id=f><td>b</td></tr></table>` with `tbody` dropped is one table again, browser-tree-equal to the input. |
| 8 | Tests pinned the regression | `462adc5` | `testTableFormDoesNotContainFollowingContent` and `testDroppedTableModeForm` now assert the exact browser tree of the output (form childless, table split around the fostered text) and that the text of input and output is identical, so nothing is lost or reordered. |

### On finding 7, what stays

`<table><form>y<tr><td>x` still comes out as `<table><form></form></table>y<table>…`, and
`<table><form id=f><b>q</b></table>` as `<table><form id="f"></form></table><b>q</b>`. That is
#481's design for anything a browser foster-parents: `main` gives
`<table></table>y<table><tbody>…` for `<table>y<tr><td>x` and `<table></table><b>q</b>` for
`<table><b>q</b></table>`. The form path now behaves like the text path instead of splitting
where the text path would not. Changing the split itself is a #481 question, not a #484 one;
the tests document it.

### Two existing expectations changed, both to keep text

- `testTableModeFormUsesEmittedTableContext`: `<table><template><form>x` with the template
  dropped was pinned as `<table><form></form></table>` — the `x` deleted. Now
  `<table><form></form></table>x`.
- `testTextInADroppedCellOfAKeptRowIsDropped` (from `main`, #448) pinned text in a dropped
  cell of a kept row as deleted, because the only alternative then was writing it straight
  into the `<tr>`. It is now `…IsKeptBesideTheTable`: `<table><tbody><tr></tr></tbody></table>cell`,
  a fixed point, where a browser's foster-parenting puts it. The text still never enters the
  row, which is the property the test protected.

No other test line was removed or loosened.

### Verification

- `./mvnw clean verify` on JDK 11, 21 and 25: BUILD SUCCESS, 662 library tests (+9 over
  `19b5fd2`) and 7 example tests, JPMS consumer check included.
- Fuzzers green at seeds 7, 99, 1234, 20260918 and 424242.
- Differential probe, 4,000 hostile inputs × 8 policies plus the plain receiver at nesting
  limits 0–6, fixed build **vs `19b5fd2`: 0 regressions on every metric**; 4 idempotence and 3
  tree-stability cases fixed. **Vs `main`**: plain-receiver balance 13,188 broken streams → 0
  with 0 regressions (unchanged from the PR head); idempotence 256 fixed / 18 exposed (was 22);
  tree stability 260 fixed / 23 exposed (was 25); policy-to-renderer events 0 unbalanced.
- mXSS probe, 60,000 hostile foreign-content and raw-text cases: the only failure kind is
  reparse instability, at the same counts as `19b5fd2` (1,173 / 1,204 / 1,234 per seed); zero
  script elements, event handlers or `javascript:` URLs.
- The eight reproductions from the review comment, re-run against the fixed build, all come
  out as described above.

### Not fixed in this pass

- **Constant-factor slowdown on malformed input.** The quadratic is gone, but `69cde7e`'s
  per-tag bookkeeping still costs ~8–10× on unclosed-tag chains with the prepackaged policies
  (8,000 repeats: unclosed `<table><tr><td>` 57 ms vs 6 ms on `main`; `<b><i><a>` 47 vs 12;
  `<table><form>` 64 vs 7), ~1.4× on well-formed input. Linear, so not a DoS class, but it
  needs a profiling pass: the prepared-start protocol, the per-tag stack scans
  (`outputTableContextForStart`, `hasOpenTemplateElement`, `hasUnavailableInputTableInScope`)
  and allocations such as the `new ArrayList` in `textIsInForeignContent()` are the likely
  costs.
- **Formatting resumption placement** (review `:2096`, `:2098`). `<foo><b>x</foo><p>y</p>z`
  → `<foo><b>x</b></foo><b><p>y</p>z</b>`; `<math><b></math><li>` is not a fixed point.
  Formatting is resumed in front of a start tag that a browser does not reconstruct active
  formatting for (`p`, `hr`, `table`, `li`, …). The fix is to resume only for text and for
  start tags that reconstruct, inside the element being opened — a separate change with its own
  tests.
- **Table parts without a table** (`:837`): `<ul><th><svg>` under a policy that drops `table`
  gives `<tbody><tr><svg>` and is not a fixed point. Pre-existing: `main` gives
  `<tbody><tr></tr></tbody>` for `<th>x` under the same policy.
- **Text inside a foreign element under `ul`/`select` gets an implied `li`**
  (`<ul><math><mi>x</mi></math></ul>` → `<mi><li>x</li></mi>`). Pre-existing on `main`. Skipping
  HTML containment for foreign text is not safe as things stand: the policy's text gate treats a
  foreign `tbody` as an HTML one and would then drop `D` in `<ul><svg><tbody>D`.
- `<li>` inside `colgroup` is not foster-parented (colgroup is not a foster-parenting context);
  surfaced by finding 5's input, pre-existing.
- `:356` (package-private `instanceof` feature detection), `:807` (`pushPassthrough` when the
  policy emitted nothing) and the smaller items in the review comment are untouched.

jmanico:
## Follow-up: the constant-factor slowdown, two small items, and a correction

Three more commits on top of `462adc5`, ending at `f043355`.

| Commit | What |
|---|---|
| `51e5a35` | **Performance.** Profiled the fixed branch with JFR on the malformed shapes. Seventy percent of samples on the unclosed-`td` chain were one method, `outputAllowsImplicitTableReturn()`, a full stack scan run for every table part *before* the cheap flags that decide whether its answer matters; another tenth was `suppressedMappedForeignTableIndex()`, scanned on every event. Both, and two siblings, only return a non-trivial answer when a bit in `outputlessTablesWithEmittedParts` or `suppressedMappedForeignSubtrees` is set, which nothing does unless a policy drops or renames a table. They now return at once when the bitset is empty, and the `prepareForContent` condition tests the flags before the scan. Every reordered conjunct is side-effect free: the differential probe (4,000 inputs × 8 policies + the plain receiver at 7 limits, 60,000 case lines) is **byte-identical** before and after. |
| `adf82d8` | A form start the balancer ignores because the form pointer is set is now reported to `HtmlChangeListener`, as tags dropped at the nesting limit already were (`<form id=a><form id=b>x` → log `<form>`). |
| `f043355` | `reopenedTableWasRenamed` is reset in `openDocument()`/`closeDocument()`, and `droppedSuppressedOptionPassthroughDepth` with the rest of the suppression state. Neither could be read stale as the code stands; hygiene. |

### Throughput, prepackaged `Sanitizers`, 8,000 repeats

| shape | `main` | before `51e5a35` | after |
|---|---|---|---|
| unclosed `<table><tr><td>x` | 5.9 ms | 57.3 ms | **9.2 ms** |
| unclosed `<table><form>w<tr><td>v` | 7.3 ms | 63.8 ms | **12.0 ms** |
| unclosed `<b><i><a href=/x>q</zzz>` | 12.3 ms | 47.2 ms | 29.4 ms |
| well-formed table / prose / flat rows | 4.6 / 7.1 / 7.2 ms | 6.8 / 8.6 / 10.0 ms | 6.1 / 8.2 / 9.6 ms |

What remains on the `b/i` chain (~2.2× main) is two scans that need real bookkeeping to remove:
`outputTableContextForStart()` ending in the policy's `formStartTagUsesTableRules()` (new in this PR,
~20% of samples) and `hasOpenLinkInFormattingScope()` (on `main` too, ~20%, only for `<a>`). Left as is.

### Correction to my earlier comment

I called the formatting-resumption placement (`:2096`, `:2098`) a regression against `main`. It is
not. `main` gives `<div><b>x</div><p>y</p>z` → `<div><b>x</b></div><b><p>y</p>z</b>` and is not a
fixed point on `<li><b>x</li><li>y` either, and the old balancer test pins
`<b><b><p>x</p></b></b>`. This PR only makes the same shape reachable through an unrecognized end tag
closing formatting, which is browser-correct. Where formatting resumes relative to a block start is a
long-standing question for a separate change.

### Verification

- `./mvnw clean verify` on JDK 11, 21 and 25: BUILD SUCCESS, 663 library tests and 7 example
  tests, JPMS consumer check included.
- Fuzzers green at seeds 7, 99, 1234, 20260918, 424242 and 2026.
- Differential probe byte-identical across the performance commit; the reporting and reset commits
  change no output.

jmanico:
## A second review found a blocking regression in the fix series; fixed in `f2fc154`

### The regression

`f3f4406` made forwarded elements count toward the nesting limit. That bounded the list an end
tag scans, which fixed the quadratic — but it charged the output-depth budget for elements that
nest nothing. A policy drops most unknown tags, so a run of them exhausted the budget and
everything after it was silently deleted:

```
Sanitizers.BLOCKS+FORMATTING+LINKS+TABLES        main / 19b5fd2              f043355
255x <o:p> + <p>Hello <b>world</b></p>           <p>Hello <b>world</b></p>   <p>Hello world</p>
256x  "                                          <p>Hello <b>world</b></p>   Hello world
256x <my-widget> + para + link + table           full markup                 paralinkc
```

`<o:p>` is what a pasted word-processor document opens; closed pairs were fine, an unclosed run
was not. Both `main` and the branch before this series kept all of it. This was a real
content-deletion regression and it would have shipped.

### The fix

The two concerns are now separate:

- **Lookup.** Forwarded elements are indexed by canonical name, so an end tag finds the one it
  closes in constant time however many are open. The quadratic is fixed by the index rather than
  by a bound — and is faster for it: 32,000 repeats of `"<foo></div>"` go from 20.6 ms to
  **11.1 ms** (was 5,152 ms before the series).
- **Budget.** `effectiveNestingDepth()` counts what actually nests again: behind a policy, the
  policy's own output depth, which already covers the forwarded elements it emitted and rightly
  ignores the ones it dropped; with no policy to ask, every forwarded element reached the
  receiver below, so those still count — the limit remains a real bound for unrecognized names
  on a plain receiver, which it was **not** on `main`.

The redundant `passthroughCanonNames` list goes with it (every push site already passes a
canonical name). `testRunOfUnrecognizedTagsIsLinear` asserted only that the result was empty,
which this regression also produced; both its cases now keep a control suffix, so it fails on
the budget bug as well as on the quadratic. Both new tests fail on `f043355` for the stated
reasons.

### Correction to the review comment above

Three of that review's findings were checked against the PR base only. Against `main` they read
differently, and I verified each:

- **Implied `<li>` inside a foreign root / namespace flip.** The `<style>` example does not
  reproduce. For `<textarea>`, `<a>` and `<mi>`, **head matches `main` exactly**; the PR base was
  the outlier, and `350cc3a` restored `main`'s behaviour rather than breaking it. The namespace
  oddity is real but pre-existing.
- **Orphan `<caption>` under `Sanitizers.TABLES`.** Same shape: **head == `main`**, base was
  better. `350cc3a` gave up an improvement rather than introducing a regression.
- **`disallowTextIn` bypass.** The base already emits the text for `disallowTextIn("form")`,
  `("tbody")` and `("table")`, so "base suppresses, head emits" is wrong. Head does newly bypass
  the `("tr")` case. Against `main` all of them are bypassed — a real policy-enforcement
  regression, mostly from the earlier commits in this PR and slightly widened by `2c2d2bf`.

The remaining findings in that comment are not independently verified here.

### Verification of `f2fc154`

- `./mvnw clean verify` on JDK 11, 21, 25: BUILD SUCCESS, **664** library tests + 7 example
  tests, JPMS check included. Fuzzers green at seeds 7, 99, 1234, 20260918, 424242, 2026.
- Differential probe (4,000 hostile inputs × 8 policies + plain receiver at limits 0–6):
  **identical to `f043355` on every metric**; vs `19b5fd2` **0 regressions** and 4 idem + 3 tree
  fixed; vs `main` 256 idem / 260 tree / 13,188 event-balance fixes with 18 / 23 / 0 regressions.
- mXSS, 60,000 hostile foreign and raw-text cases: only reparse instability, unchanged counts,
  **zero** script elements, event handlers or `javascript:` URLs.
- Output nesting depth stays within the limit for every probed shape at limits 0, 1, 2, 3, 5, 8.

### Still open, for a follow-up issue

Not regressions from this series — all three are on `main` today:

- `Sanitizers.TABLES` on `<template><ul><math><caption>TEXT` emits an orphan `<caption>` that is
  not a fixed point and whose text a browser drops entirely.
- `<ul><svg><textarea>` puts the implied `<li>` inside the foreign root and flips `textarea` to
  the HTML namespace on reparse.
- `disallowTextIn` is bypassed for text foster-parented out of a retired table.

Plus two documentation items on this series: `setNestingLimit()` throws `IllegalStateException`
in more cases than before with a one-line javadoc that documents no exception, and the new
`discardedTag` notification for pointer-ignored forms is not in the change log.

jmanico:
## Third pass: the fix series itself, every finding checked against `main`

Of the second review's first four findings checked, three had been measured against the PR base
only and pointed the wrong way against `main`. So every one of its fifteen was run here on four
builds — `main` (`569d463`), the PR base (`19b5fd2`), the pre-fix head (`f043355`) and the current
head — before being classified. Commit `33c2b90` fixes the two that hold up.

| # | Finding | Verdict |
|---|---|---|
| 1 | implied `<li>` inside a foreign root, namespace flip | **head == `main`**; the PR base was the outlier. Pre-existing. |
| 2 | dropped unknown tags exhaust the nesting budget | **real, from `f3f4406`**; fixed in `f2fc154` |
| 3 | `</form>` inside foreign output while the tracker is unknown latches the pointer | **real, from `58dd112`**; fixed in `33c2b90` |
| 4 | orphan `<caption>` under `Sanitizers.TABLES` | **head == `main`**; base was better. Pre-existing. |
| 5 | dropped `thead`/`tfoot` exempted from retirement | **real, from `147f3fa`** (mine); `main` has the same instability, the base had fixed it. Fixed in `33c2b90` |
| 6 | orphan / self-nested `<li>` after a retired table | **not reproduced** — identical, idempotent output on all four builds |
| 7 | `disallowTextIn` bypassed beside a retired table | base already bypasses `form`/`tbody`/`table`; head adds `tr`. Against `main` all bypassed — from the earlier #484 commits, widened slightly by `2c2d2bf`. **Open, pre-existing in the PR** |
| 8 | raw name at the option-suppression site | **not reproducible** — converted in `756d10c`, grep finds no raw-name forward |
| 9 | `openDocument()` does not reset the stacks | **pre-existing on `main`**, which also leaves the fourteen stacks to `closeDocument()`; `main`'s second document is polluted the same way |
| 10 | `<select><svg>x</svg></select>` grows per pass | **pre-existing on `main`**: 38→47→56→65 on all four builds, only the wrapper's position differs |
| 11 | a form the ElementPolicy rejects latches the pointer | **correct behaviour**: a browser ignores the second `<form>` because the first set its pointer, so `ac` is the sanitization of the input's DOM; `main`'s extra form was the error |
| 12 | two new tests do not pin their fixes | **confirmed**; linearity test fixed in `f2fc154`, the form test now checks the depth-255/256 boundary in `33c2b90` |
| 13 | reported name raw vs canonical | **confirmed, minor**: six sites pass the name as received, two the canonical one; only a recasing preprocessor can tell them apart. Follow-up |
| 14 | `closeDocument()` reset asymmetry | balancer side is hygiene, pre-existing pattern; the policy-side claim (`outputFormElementPointerIsSet()` reads true after close) **does not reproduce** |
| 15 | `passthroughCanonNames` redundant | fixed in `f2fc154` |

### The two fixes in `33c2b90`

**F3.** The rule from `58dd112` — follow the output namespace once the tracker is unknown — was
applied afresh at the end tag. A form in an HTML integration point sets the pointer under HTML
rules; after a stray `</tr>` its `</form>` arrives while the output is still inside the svg, was
called foreign, and never cleared the pointer, so every later form was dropped. The end tag now
replays the start's decision: while the form that holds the pointer is open, its end tag closes
it under HTML rules whatever the output namespace is now.

```
<svg><foreignObject><form>a</tr></form></foreignObject></svg><form>b</form>
  main / base:  …</svg><form>b</form>      f043355 / f2fc154:  …</svg>b      now:  …</svg><form>b</form>
```

**F5.** The exemption now covers only a dropped `tbody` or `tr`, which is exactly what the output
parser implies again. `<table><thead><form><tr><td>x` with thead dropped and tbody kept is a fixed
point again, as it was on the base.

### Verification of `33c2b90`

- `./mvnw clean verify` on JDK 11, 21, 25: BUILD SUCCESS, 666 library tests + 7 example tests,
  JPMS check. Fuzzers green at six seeds.
- Both new tests fail on `f2fc154` for the stated reasons.
- Differential probe (4,000 inputs × 8 policies + plain receiver at limits 0–6): **identical to
  `f2fc154` on every metric**; vs `19b5fd2` 0 regressions, 4 idem + 3 tree fixed; vs `main`
  256 / 260 / 13,188 fixes with 18 / 23 / 0 regressions, unchanged since the series began.
- F2 still fixed; `"<foo></div>" x 32000` in 11.3 ms; mXSS probe unchanged with nothing
  script-capable.

The automated review of `f2fc154` itself was cut off by the org spend limit after confirming that
every mutation of `passthroughNames` is LIFO. I read that commit directly: the per-name index is
kept consistent by push, `popPassthrough` and `clear`, and the policy branch of
`effectiveNestingDepth()` is byte-for-byte the base's, which is why the probe was identical.

### Still open

Split as agreed:

- https://github.com/OWASP/java-html-sanitizer/issues/491 — follow-up for this series: the `setNestingLimit()` javadoc, the change log line for
  the new `discardedTag` notification, the raw-vs-canonical report name, and the
  `closeDocument()` reset asymmetry. One commit, one PR, after this merges.
- https://github.com/OWASP/java-html-sanitizer/issues/492 — the five behaviours that are on `main` today (findings 1, 4, 7, 9, 10), with
  reproductions.

jmanico:
## Third pass on `33c2b90` vs `main` @ `569d463` — four new regressions, please do not merge yet

This pass used a **new** differential probe, not the corpus from the earlier comments, so it
covers different ground: 34 policies including recasing pre/postprocessors, `disallowTextIn`
gates, policies that drop `select`, `svg`/`math` or `table` while keeping custom elements, and
shapes inside `<svg><foreignObject>`. Both builds were compiled from clean worktrees and every
case below was re-run by hand on both. A case counts as a regression only when `main` had **no**
flag at all on it, so `main`'s own instabilities do not inflate the numbers.

| Check | Result |
|---|---|
| Sanitize cases (15,430 inputs × 34 policies) | 524,620 per build |
| Plain `TagBalancingHtmlStreamEventReceiver`, LIFO-checked, limits 0–8 | 36,063 per build |
| Exceptions out of `sanitize()`, unbalanced policy→renderer streams, allowlist widening, listener parity | **0** on head |
| Script elements, `on*` attributes, `javascript:` URLs under script-free policies | **0** on both |
| Output depth over the nesting limit on the plain receiver | 0 on head (3 on `main`) |
| Merge of `33c2b90` into `main`, `./mvnw verify` | BUILD SUCCESS, 666 library tests |
| Scaling, 40 shapes × 3 policies, N = 2k…32k | linear on head; no quadratic |

Strict regressions: 1,979 (policy, input) pairs out of 524,620. They reduce to the four shapes
below plus classes already filed in #492.

### 1. A dropped `select` wraps its contents *and everything after it* in a stray `<li>`

Any policy that drops `select` but allows `li`, which `Sanitizers.BLOCKS` does. Dropping only
`option` is fine; dropping `select` triggers it. On `main` the text comes out bare.

```
Sanitizers.BLOCKS.and(Sanitizers.FORMATTING)
<select><option>x</option></select><p>y</p><p>z</p>
  main  x<p>y</p><p>z</p>
  head  <li>x<p>y</p><p>z</p></li>          second pass: <ul><li>x<p>y</p><p>z</p></li></ul>
<form><select><option>x</option></select></form><p>y</p>
  main  x<p>y</p>                            head  <li>x</li><p>y</p>
<p>a</p><select><option>x</option></select><p>y</p>
  main  <p>a</p>x<p>y</p>                    head  <p>a</p><li>x<p>y</p></li>
```

`<select>x` already gave `<li>x</li>` on `main` (the implied-wrapper class in #492), but this
branch extends it to the ordinary `<select><option>` shape and the item stays open until the
enclosing block closes. Not a fixed point, very common input, published policy. Blocker.

### 2. Allowed content deleted: dropped or renamed foreign root foster-parented out of a table, with a custom-element child

Policy drops `svg`/`math` (or renames them) but allows the custom element.

```
<table><tr><svg><o:p>x</o:p></svg><td>y
  main  <table><tbody><tr></tr></tbody></table><o:p>x</o:p><table>…<td>y…
  head  <table><tbody><tr></tr></tbody></table><table>…<td>y…          x and <o:p> gone
<table><svg><o:p>tail
  main  <table></table><o:p>tail</o:p>       head  <table></table>
<table><tr><svg><foo>x</foo></svg><td>y  /  <table><tr><math><foo>x</foo></math><td>y  /
<table><tfoot><svg><o:p>tail</o:p></svg><tr><td>y  /  <table><colgroup><svg><foo>x</foo></svg><col><tr><td>y
  same deletion; svg→div rename gives <div></div>
```

A `<b>` child instead of a custom element is fine, and the prepackaged policies are unaffected
because they drop the custom element too. Mechanism:
`TagBalancingHtmlStreamEventReceiver.openTag`, `suppressForeignBreakoutBesidePushedTable`
(computed at `:736`, applied at `:851`) is true because the *output* has no foreign root — the
policy dropped it — so the child goes through `openTagWithoutOutputOrContent` and its whole
subtree is suppressed. Introduced in `69cde7e`.

### 3. Allowed content deleted: table with a form inside an integration point, `table` dropped or renamed

```
policy drops table, keeps form/tr/td
<svg><foreignObject><table><form><tr><td>y</td></tr></table>
  main  <svg><foreignObject><form><tbody><tr><td>y</td></tr></tbody></form></foreignObject></svg>
  head  <svg><foreignObject><form></form></foreignObject></svg>
<svg><foreignObject><table><form><tr><td>y</td></tr></table></foreignObject></svg><p>after</p>
  head  …<form></form></foreignObject></svg><p>after</p>                    y gone, after kept
```

Same under `<svg><desc>` and `<math><mtext>`, with `<tbody><form>` and `<tr><form>`, and with
`table` renamed to `div`. Outside an integration point (`<div><table><form><tr><td>y`) it is
correct on both. Prepackaged policies are unaffected only because they drop `tr`/`td` as well.

### 4. Formatting closed by a foreign end tag is resumed inside the implied `<ul>`, before its `<li>`

```
Sanitizers.BLOCKS.and(Sanitizers.FORMATTING)
<math><b>x</math><li>y</li>
  main  <b>x</b><ul><li>y</li></ul>
  head  <b>x</b><ul><b><ul><li>y</li></ul></b></ul>     second pass adds another <li> level
<svg><b></svg><li>x     same shape under an allow-all policy
```

The resumption itself is new in this branch, so this is not the pre-existing
`<div><b>x</div><p>` placement question from the second comment: `main` is clean and idempotent
on these inputs. The resumed element is emitted after the wrapper is implied and before the
tag's own containment check, which then implies the wrapper a second time.

### The eight first-round blockers hold

Re-run on `33c2b90`: `"<foo></div>" x 32000` in 11 ms; `sanitize(in)` and
`sanitize(in, listener, ctx)` byte-identical on all 524,620 cases; no `IllegalStateException`
anywhere; `<tbody><form>B` keeps `B`; `<select><svg>` is a fixed point; forms after a
depth-256 dropped form are emitted; `<svg><foreignObject><form>a</tr></form></foreignObject></svg><form>b</form>`
keeps both forms; 255/256/300 unclosed `<o:p>` before `<p>Hello <b>world</b></p>` keep the markup.

### Smaller notes

- **Performance.** `main` has a quadratic this branch fixes: `<svg>` followed by
  `"<foo></svg>" x 32000` takes 2,136 ms on `main` and 9.6 ms here. Constant factors on
  malformed input under drop-heavy policies are still high, all linear:

  | shape × 32,000 | policy | `main` | head |
  |---|---|---|---|
  | `<template><table><form>x` | BLOCKS+FORMATTING+LINKS+TABLES+IMAGES+STYLES | 13 ms | 257 ms |
  | `<table><tr><td>x` unclosed | `allowElements("p")` | 16 ms | 360 ms |
  | `<ul><li><foo>` | allow-all | 16 ms | 128 ms |
  | `<b><i><a href=/x>q</zzz>` | prepackaged | 41 ms | 118 ms |

- **Plain receiver.** Thousands of streams that were unbalanced on `main` are balanced here and
  the depth never exceeds the limit. Five artificial event streams newly reach the renderer's
  error callback because an unrecognized start tag is forwarded inside an open `iframe`/`xmp`;
  `main` does the same at other limits, so this is a shift, not a new class. `setNestingLimit`
  throwing when forwarded elements are open is the #491 item.
- **`disallowTextIn`.** Checking the text node's *parent* in the browser tree, only two random
  inputs out of 524k put text directly inside a disallowed element here where `main` suppressed
  it; both are foster-parented text, i.e. #492 item 3.
- **`<table><form id=a>x</table>` → `<table><form id="a"></form></table>x`** flips 354 keep-policy
  cases from browser-equal to reordered while fixing 401 others. That is the #481 design already
  discussed above; noted, not counted.
- Housekeeping: the eight tests that appear deleted in the diff were moved with identical bodies,
  the two changed expectations are documented, the change-log bullet is present, and the merge
  base is current `main`.

### Suggested order of work

1 first (user-visible under a published policy), then 2 and 3 (same bar as the earlier
text-deletion blocker), then 4. One commit each with a test that fails on `33c2b90`. Acceptance:
the probe run against `33c2b90` shows zero strict regressions and the reproductions above fixed,
and the run against `main` keeps today's fix counts. The probe, joiner and the minimised
reproductions with both builds' outputs are in `java-html-sanitizer-handoff/probes-review3/`
(local), with a README.

Nothing here is an XSS bypass and none of it needs an advisory.

jmanico:
## Fix series for the third pass: seven commits on top of `33c2b90`, ending at `6c1fb9f`

One commit per finding, each with a test that fails on the commit before it for the stated
reason (checked by reverting the source under the new test). Commits 1–4 are the four
regressions from the review comment above, in its suggested order. Commit 5 fixes a
quadratic that the acceptance run turned up and that none of the earlier timing shapes
reached. Commits 6 and 7 come from an independent adversarial review of commits 1–5, which
verified two refinements against `main`, `33c2b90` and the fixed head; the rest of its
findings were pre-existing classes.

| # | Finding | Commit | Fix |
|---|---|---|---|
| 1 | Dropped `select` leaves a stray `<li>` that swallows what follows | `eb11921` | An option whose input container is a dropped `select` or `optgroup` keeps that container. The output-aware rule sent it to the body, where the containment metadata implies a select the same policy drops again, and preparing that select under the dropped one implied the canned list item. `<select><option>x</option></select><p>y</p>` → `x<p>y</p>` again under `Sanitizers.BLOCKS`. |
| 2 | Children of a dropped or renamed foreign root beside a table deleted | `c772698` | The breakout suppression needs the output to have entered the foreign region. Each forwarded element records whether the policy emitted it as a foreign element or root, and the forwarded roots are indexed, so the check is constant time. `<table><tr><svg><o:p>x</o:p></svg><td>y` keeps `<o:p>x</o:p>`; the renamed-template case the existing test pins is unchanged. |
| 3 | Cell text of a dropped table inside an integration point deleted | `cf8bad6` | The policy's fail-closed rule for text in a suppressed table part, written for the renderer's lexical SVG left open after an emitted breakout, also fired inside an integration point. It fails closed only when the output tracker has no foreign root left. `<svg><foreignObject><table><form><tr><td>y` → `<form>y</form>`. |
| 4 | Formatting resumed inside the list implied for a list item | `3387e8c` | A formatting element is resumed for a tag only if it can hold the tag with nothing implied between them; otherwise it stays queued for the content inside the tag, where a browser reconstructs it. `<math><b>x</math><li>y</li>` → `<b>x</b><ul><li><b>y</b></li></ul>`, the browser's tree; the same rule makes `<ul><li><b>x</li><li>y` and `<ol><li><b>x</li><li>y</li></ol>` fixed points, which they were not on `main` either. |
| 5 | **New:** a run of unclosed unknown tags followed by ordinary markup was quadratic | `cfb21b6` | The policy keeps a logical entry per dropped element, unbounded by the nesting limit, and its table-scope check for form starts walked that stack for every recognized start tag. `"<foo>" x 128,000` then `"<li>" x 128,000` under `BLOCKS+FORMATTING+LINKS+TABLES`: `main` 42 ms, `33c2b90` **11.7 s**, now under 100 ms; `<p>`, `<table>` and `<option>` runs the same. The six names that decide the check are indexed as they are pushed. Pure optimisation: the probe's output is byte-identical before and after. |
| 6 | `disallowTextIn("form")` bypassed by a suppressed cell in an integration point (from the independent pass) | `bd1bf23` | Commit 3 let the suppressed cell's own gate admit `y` into a form the policy disallows text in, and the next pass removed it. The gate now also keeps out text the policy disallows in the element the text lands in, so `<form></form>` comes out at once. The part's own gate still applies: a table renamed to `div` keeps the text of its suppressed cells, and `disallowTextIn("td")` drops it, as on `main`. |
| 7 | Formatting resumed after its own end tag, and a link after the next link (from the independent pass) | `6c1fb9f` | `<math><b>x</math></b><li>y` bolded `y` and `<foo><a href=x>x</foo><li><a href=y>y</a>z` linked `z`. A browser drops such an element from its active-formatting list when its end tag arrives with the element no longer open, and drops an earlier link when the next starts; the queue now forgets them too. `testPushedOutLinkIsNotResumedAroundAnotherLink` pinned the pushed-out link resumed around the text after the second link; its expectation now follows the browser, and its point, no link written around or inside another, is unchanged. |

A first attempt at 4 moved the resumption earlier; it broke the existing nested-link test and
was replaced before anything was pushed. A first attempt at 6 made a suppressed part inherit
the gate in force, which deleted the text of every cell under a table renamed to `div`
(1,423 probe cases); the probe caught it before it was pushed.

### Verification of `6c1fb9f`

- `./mvnw clean verify` on the branch: BUILD SUCCESS, 673 library tests and 7 example tests,
  JPMS check included. Fuzzers green at seeds 7, 484, 20250912, 424242 and 2026. The merge of
  `6c1fb9f` into `main` (scratch worktree) builds and passes the same suite.
- Differential probe, the corpus of the third-pass comment (524,620 sanitize cases across 34
  policies, 36,063 plain-receiver runs at limits 0–8), counting a case only where the baseline
  had no flag at all:
  - **vs `main`: 1,297 regressions on `33c2b90` → 682 now.** Two are new: one random input,
    under the table-dropping and the table-renaming policy, where a stale `<b>` used to be
    resumed and its open reset the text gate, so `tail` slipped past the policy's deliberate
    fail-closed rule for text in a suppressed part after an emitted breakout; with 7 the rule
    applies and the text is dropped, as it is for the shape listed below. Cases flagged on
    `main` and clean here: 37,661 → 38,714.
  - **vs `33c2b90`: 1,696 cases fixed, 28 newly flagged**, none a new class. Twenty are an
    `<option>` or `<optgroup>` under a dropped table cell that gets no `<select>` on the first
    pass (`<table><th><option>tail` → `<option>tail</option>`, not a fixed point on `main` or
    `33c2b90` either); a queued formatting element used to be resumed around the option and the
    select was implied inside it, which hid the gap. Five are under the svg→div rename policy,
    where content `33c2b90` deleted is now kept. Two are the fail-closed case above, one is the
    pre-existing `<ul>` inside `<p>` quirk exposed once a stale `<strong>` is no longer resumed.
  - Plain receiver: zero regressions vs `33c2b90`, depth never over the limit; 19 runs on one
    input family changed because a stale `nobr` is no longer resumed after its own end tag.
  - 205 hand-minimised cases from the review: 39 text-loss fixes, 35 idempotence fixes, no case
    worse than either baseline.
- Timing: no shape above 2.2x per doubling; median cost unchanged from `33c2b90`. One constant
  factor grew: `<select><option><foo>` x 32,000 under the prepackaged policies, 20.6 → 110 ms,
  because the dropped pairs now nest to the limit and each event scans the full balancer stack.
  Linear.

### Still open, none introduced by this series

- Text inside an HTML breakout element that sits lexically inside a foreign subtree with an
  intervening foreign element is dropped by the policy's fail-closed rule:
  `<svg><tbody><table><foreignObject><form>tail` loses `tail` (`main` keeps it, `33c2b90`
  loses it). The rule is deliberate; whether it should apply there is a maintainer call.
- A second, pointer-ignored form in a row of a dropped table inside an integration point:
  `<svg><foreignObject><table><form id=a><tr><form id=r><td>y</td></tr></table>z` keeps `y`
  now but still loses `z` (`main` keeps both).
- `<math><b>x</math><li>y<li>z` nests the second item inside the resumed `<b>`, as
  `<ul><li><b>x<li>y` does on `main`: the resumed element hides the open item from the next
  `<li>`. Stable output.
- The 682 remaining differences from `main` are the implied `<li>`/`<select>` wrapper class of
  #492, recasing-preprocessor shapes, `<style>` text after an emitted breakout out of a
  policy-created foreign root, and a renamed form inside an integration point. Reproductions
  are in the local probe folder.

Nothing here needs an advisory; none of it is an XSS bypass.

Issue #492, opened 2026-09-18
Tag balancer: five foster-parenting and text-gate gaps present on main


jmanico:
Five balancer behaviours found while reviewing #490 that are present on `main` (`569d463`) today
and are not changed by that PR. Each was reproduced on `main`, the #490 base, and the #490 head;
none is an XSS, and none needs an advisory. They are here so they are not lost.

1. **An implied `<li>` lands inside a foreign root and flips a child's namespace.** With a policy
   allowing `ul`, `li`, `svg`, `textarea`: `<ul><svg><textarea>x</textarea></svg></ul>` →
   `<ul><svg><li><textarea>x</textarea></li></svg></ul>`. The browser's tree of that output is
   `ul > {SVG}svg > li > textarea` — `li` is an HTML breakout name, so the `textarea` becomes an
   HTML `textarea` rather than the SVG element the input had. Same with `<a>` and `<mi>`. The
   containment metadata has no entry for foreign content, and text inside a foreign element under
   a `ul` still gets the `li` the metadata implies for `ul`. Skipping HTML containment for foreign
   text is not safe as things stand: the policy's text gate treats a foreign `tbody` as an HTML
   one and would drop `D` in `<ul><svg><tbody>D`.

2. **`Sanitizers.TABLES` emits an orphan table part that is not a fixed point.**
   `<template><ul><math><caption>TEXT` → `<caption>TEXT</caption>`; a second pass gives
   `<table><caption>TEXT</caption></table>`, and a browser parsing the first output drops the
   `caption` element entirely, leaving bare text. Same with `<svg>` for `<math>` and `<colgroup>`
   for `<caption>`. A published prepackaged policy.

3. **`disallowTextIn` is bypassed for text foster-parented out of a retired table.**
   `new HtmlPolicyBuilder().allowElements("table","tbody","tr","td","form").disallowTextIn("form")`
   on `<tbody><form>B` emits `B` (beside the table). The gate follows the nearest kept element,
   and when the table's output is retired for the text, the element the author disallowed text in
   is no longer the one judged. `main` suppresses it only because it never retires the table
   there and keeps the text inside the form, which is the #484 bug.

4. **`openDocument()` does not reset the stacks; only `closeDocument()` does.** A caller whose
   receiver throws mid-document never reaches `closeDocument()`; the next `openDocument()` then
   starts with the previous document's `openElements` and the fourteen parallel structures, and
   the second document's output begins with the first document's close tags:
[...]

PR #498, opened 2026-09-19
Judge text after a form popped in an implied table by the form's text gate

jmanico:
**Draft: not ready to merge.**  A `/code-review` run reproduced text loss and
a quadratic scan in the gate as this branch holds it: the gate does not end
where the javadoc says it does, so with `disallowTextIn("form")` a single
`<p><tbody><form>x</p>` deletes every later bare text run in the document,
and a document of dropped tags costs 2.7 s where main costs 55 ms.  Notes for
whoever picks this up, including the direction the review suggests, are in
the handoff folder, `issue492-reviews/`.

Item 3 of #492: `disallowTextIn` is bypassed for text foster-parented out of a retired table.

## What broke

`new HtmlPolicyBuilder().allowElements("table","tbody","tr","td","form").disallowTextIn("form")` on `<tbody><form>B` gave `<table><tbody><form></form></tbody></table>B`: the text the author disallowed in the form came out beside the table. A browser reading that input ignores the stray `tbody`, opens the form, and puts `B` in it, so the text is the form's. Same for `<tr><form>B`, `<colgroup><form>B`, and with dropped elements between (`<tbody><form><p>B` with `p` not allowed).

## Why

The tag balancer implies a table for a table part the input writes without one. Inside that table the form start takes the in-table rule of #484: the form is inserted and popped at once, written as `<form></form>`, and the text after it is foster-parented out of the table. The policy's text gate follows the nearest kept element, which is now whatever holds the table, so the gate the author set on `form` no longer judges the text. `main` before #484 kept the text inside the form, which was the #484 bug; the gate held only by accident.

For a table the input has, `<table><form>B`, a browser does the same as the sanitizer: the form is popped and `B` lands in front of the table, outside the form, so keeping it is right there and nothing changes.

## What changed

- `TagBalancingHtmlStreamEventReceiver`: when the in-table rule pops an HTML form in a table the balancer implied (`impliedInputTables`), outside template contents, the balancer tells the policy to hold the form's text gate (`FormPointerPolicy.holdRetiredFormGate`, with whether the policy emitted the form), before the table's output is retired for the form, and tells it to release the gate (`releaseRetiredFormGate`) at every form end tag, under HTML and under foreign rules. Both are default methods of the interface, forwarded by the change reporter. The balancer's stack and its foster-parenting are unchanged: content is still written beside the table.
- `ElementAndAttributePolicyBasedSanitizerPolicy`: holding the gate does nothing when the form may hold text: for a kept form, unless the policy disallows text in it; for a dropped form, unless the author wrote `disallowTextIn("form")`, the rule any dropped element's text follows. So nothing changes for other policies. When it holds, the policy remembers the depth of the implied table's entry, found through its indexed table-scope entries, and the depth the gate was held at; text is the form's, and stays out, while it arrives at or above the held depth and no further down than the table's level, which is where the balancer writes what it foster-parents out of the table. An element opened after the hold holds its own text; text in elements the policy dropped is the form's, as text in a dropped element is judged by the nearest emitted element anyway. One helper lowers the held depth as the stack shrinks, when the table is pushed out, and ends the gate when the table's container closes; the release ends it too. The gate never touches `skipText`, so nothing is restored afterwards. No discarded tag is reported.
- Tests in `HtmlSanitizerTest`: the issue's shape and its neighbours (`tr`, `colgroup`, `thead`, `tfoot`, `caption`, `th`, a dropped element between, text after `</form>` and after `</tbody>`, a cell, a table the input has, hostile payloads), the same shapes with text allowed in the form, a kept `b` inside the form holding its own text with and without `</form>` and text after the form, a dropped form under `disallowTextIn("form")`, a kept `template`, a `</form>` inside `svg`, and the change listener seeing no discarded tag.
- `change_log.md` bullet.

## Before and after

Policy: `table`, `tbody`, `tr`, `td`, `form`, with `disallowTextIn("form")`.

| Input | Browser tree of input | main | This branch |
| --- | --- | --- | --- |
| `<tbody><form>B` | `form["B"]` | `<table><tbody><form></form></tbody></table>B` | `<table><tbody><form></form></tbody></table>` |
| `<tr><form>B` | `form["B"]` | `<table><tbody><tr><form></form></tr></tbody></table>B` | `<table><tbody><tr><form></form></tr></tbody></table>` |
| `<tbody><form>B</form>C` | `form["B"] "C"` | `...</table>BC` | `...</table>C` |
| `<tbody><form>B</tbody>C` | `form["BC"]` | `...</table>BC` | `...</table>` |
| `<tbody><form><p>B` (`p` dropped) | `form[p["B"]]` | `...</table>B` | `...</table>` |
| `<table><form>B` | `"B" table[form[]]` | `<table><form></form></table>B` | unchanged |
| `<td><form>B` | `form["B"]` | `<table><tbody><tr><td><form></form></td></tr></tbody></table>` | unchanged |

With text allowed in the form, `<tbody><form>B` still gives `<table><tbody><form></form></tbody></table>B`, and `<tbody><form><b>B</b>C` gives `...</table><b>B</b>C`; with the gate closed and `b` allowed, `...</table><b>B</b>`, since `B` is the `b`'s text and `C` the form's. Every output is a fixed point.

## Verification

- `./mvnw -o -ntp -B clean verify` on the final commit `d509dc3`, with main (#497 merged) merged in: build success, 687 tests, 0 failures (686 existing, unchanged, plus 1 new).
- Differential probe (`probes-review3/ReviewProbe`, 15,430 inputs x 44 policies, `san` and `events` modes, main `2bc430e` against the final commit): `san` 678,920 rows compared, 85 differ in output, no row gains or loses a flag, no row throws; `events` 36,063 rows, identical. The probe's policies disallow text in `form` only where they also keep it, and its random inputs rarely put a form directly in a stray part, so the difference is small; the tests carry the shapes. A run of the first policy-side version with the gate held after the table's retirement lost text under the policy that drops `form` on 76 rows, which the probe caught: a dropped form now holds the gate only under an explicit rule.

## Review

`/code-review` on the first policy-side version found the release restoring a stale `skipText` after the held depth had been lowered, which dropped text after the form's end tag and beyond; the hold running after the table's output had been retired for the form, so a dropped `thead`, `tfoot`, `caption` or `th` between the implied table and the form hid the table from the policy and the bypass stayed; the hold skipped for a dropped form although `disallowTextIn` applies to dropped elements; the hold applied in template contents, where a browser ignores the form start, and never released there; no release for a `</form>` under foreign rules; a full-stack walk where indexed entries exist; the end-of-gate rule in two places; and an unused parameter. All fixed above, each with a test. The probe then showed the dropped-form rule too wide, which is now the author's explicit rule only. A first version that kept a logical form entry on the balancer's stack lost text on 320 rows under policies that allow text in forms and nested a row in a row on one input; the probe caught it and the entry is gone.

## Left out

- `<tbody><form><tr><td>x</td></tr>B`: `x` in the cell is the cell's text and kept, though a browser, which ignores the row and cell too, has it in the form; `B` after the row is the form's and stays out. The cell's table structure is the sanitizer's normalisation of a stray part, applied before this item.
- A `</form>` that arrives while an element opened after the form is still open releases the gate for the text after that element closes as well, though a browser closes that element with the form; the text inside it is its own either way.
- Only a table the balancer implied takes this path. In a table the input has, the text is not the form's for a browser either.

Part of #492 (item 3)


jmanico:
Closing this draft without merging it. The implementation's text gate is over-broad and introduces the documented quadratic behavior, so green CI does not make it suitable to land. Issue #492 item 3 remains open and unresolved; closing this flawed implementation does not close the underlying gap.

